PR #4, opened 2023-03-28
feat: create a PoC based on actix actors

itegulov:


src/main.rs
```diff
@@ -1,3 +1,117 @@
...
+    );
+
+    let (actor, handle) = Actor::spawn(None, server, ())
+        .await
+        .expect("Failed to start NodeServer A");
+
+    let (node_actor, node_handle) =
+        Actor::spawn(None, actor::NodeActor, (node_id, pk_set.clone(), sk_share))
+            .await
+            .expect("Ping pong actor failed to start up!");
```
volovyks, line 54: I guess it's something from an example :)

src/actor.rs
```diff
@@ -0,0 +1,196 @@
...
+        serde_json::to_vec(&self).unwrap()
+    }
+
+    fn from_bytes(bytes: Vec<u8>) -> Self {
+        serde_json::from_slice(&bytes).unwrap()
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct SignatureResponse {
```
volovyks, line 32: It's not clear from the name how is it different from the SignResponce. Let's brainstorm on it or add some comments.

src/actor.rs
```diff
@@ -0,0 +1,196 @@
...
+}
+
+#[async_trait::async_trait]
+impl Actor for NodeActor {
+    // An actor has a message type
+    type Msg = NodeMessage;
+    // and (optionally) internal state
+    type State = NodeActorState;
+    // Startup initialization args
+    type Arguments = (NodeId, PublicKeySet, SecretKeyShare);
```
volovyks, line 69: Why do we need to declare these types here?

src/actor.rs
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    // This is our main message handler
+    async fn handle(
+        &self,
+        _myself: ActorRef<Self>,
+        message: Self::Msg,
+        state: &mut Self::State,
+    ) -> Result<(), ActorProcessingErr> {
+        let remote_actors = ractor::pg::get_members(&MPC_RECOVERY_GROUP.to_string())
```
volovyks, line 92: Do we need to fetch all the actors in each `handle()` execution?

src/actor.rs
itegulov: Yeah the names are far from perfect for sure. `SignatureResponse` is when you have the full signature ready and can respond to `NewRequest`. `SignResponse` is when you want to respond with your signature share.

src/actor.rs
itegulov: It is just a part of how ractor works. `Arguments` is what you need to pass to initialize the actor. In our case it is node id, public key set and the secret key share associated with this node.

src/actor.rs
itegulov: If we want the actor system to be dynamically joinable, then yes. But as I said we could also just get rid of actors entirely and do `N` HTTP requests to static addresses instead.

volovyks: It can be dinamically joinable in the future, but I believe it's fixed for now.
Ok, let's keep it this way. I guess there is no harm from this design at the initial stage.

PR #19, opened 2023-04-04
feat: add tracing-based logging

itegulov:
latest-screenshot


volovyks (approved):
Nice, should be super helpfull with docker problem.

PR #24, opened 2023-04-05
OAuth mock

volovyks:
I will add Google impl in a separate PR. Here is the skeleton to avoid conflicts with actor deletion.


volovyks:
Ok, look like I was 1 hour late :)

PR #313, opened 2023-10-09
feat: multichain key generation and resharing

itegulov:
This PR creates a separate crate that implement key generation and resharing based on [cait-sith](https://github.com/LIT-Protocol/cait-sith).

### High-level overview
`contract` crate:
- provides a way for registered nodes to vote on new nodes to join or old nodes to leave among other things necessary for consensus

`node` crate:
- `web` module is the communication API between nodes (to be replaced by a full connectivity mesh over TCP in the future)
- `protocol` module is the meat of potatoes of the PR containing the consensus + cryptographic layer
	* `state` contains models representing all the possible states of the node
	* `cryptography` contains a trait regulating how state can progress its cryptographic protocol (cait-sith)
	* `consensus` contains a trait regulating how state should advance according to the consensus state (fetched from the contract)
	* `message` is a way to consume and store p2p messages until they are needed for `cryptography` layer


volovyks (dismissed):
Let's merge this and iterate. Awesome work @itegulov!

volovyks (approved):

PR #320, opened 2023-10-16
Update issue templates

trechriron:
Added defect template with labels and comment-prompts for information.


itegulov:
I think this template might be better suited in the [frontend repo](https://github.com/near/fast-auth-signer)

Issue #250, opened 2023-08-09
Rate limiting


DavidM-D:
Create simple configurable limits on the number of requests that a given SDK user can make.

Aim of the game is to stop them bringing down the MPC recovery service, not that we accurately bill them.
```[tasklist]
### Tasks
- [ ] https://github.com/near/mpc-recovery/issues/186
```

volovyks:
Added on a load balancer level.

PR #321, opened 2023-10-17
Internal LB

volovyks:


kmaus-near:
@itegulov @volovyks this PR is quite large, let me give some context:

I separated dev and prod environments for ease of provisioning internal LBs for testnet and mainnet separate from the dev environment cloudrun services.

Before we merge this, the service account that runs GHA pipelines needs to have access permissions to pagoda-shared-infrastructure VPC networks, as we are adding these cloudrun services to Pagoda's internal network in order to utilize the domain names from #250  and have the Prod services only accessible via Kong API gateway (Domains).

Another update I made here is using terraform for only the initial deployment. I think in order to avoid recreating the Prod MPC cloudrun services (mainnet and testnet), we should deploy that workspace manually via local terraform (I can do this once I have the Prod variables that aren't located in the infra folder, only dev).

Once the terraform is applyed, we can then use GHA pipelines that I created in order to create new revisions VS deploying a new environment. This will be an automated process with one caveat by request from @volovyks:

 - The dev environment will only deploy the new revision if a **label** named `deploy-test` is assigned to the PR to the `develop` branch. The cloudrun service targeted will be in the `discovery-platform-dev` project, and the service name is `mpc-recovery-[leader/signer-n]-dev`
 - This means all you have to do to deploy an updated version of the dev environment is add a label to the PR called `deploy-test`

When merging to `main` branch, the GHA pipeline `deploy-prod.yml` will deploy BOTH testnet and mainnet in the same run. We can change this if need be very easily.

Please take a look at this PR and make any suggestions or call out any mistakes, if need feel free to throw a meeting invite on the calendar.

kmaus-near:
I removed the DEV GHA pipeline, let's just keep what you guys have as discussed in our call. The Prod pipeline is manually triggered for testnet/mainnet.

itegulov:
> Do we know which service account is responsible for deploying the feature environments? I need to give it access to the shared VPC networks in order for it to deploy terraform.

It should be `mpc-recovery` under the dev project

> Thinking out loud here, should we add a version tag for mainnet/testnet releases? This way we can manually roll back if any issues occur.

Yeah sounds like a good idea

volovyks (commented):
I can not technically approve this PR, because I've created it :/ So.. Approved :)

I would add a list of our envs with their proper names and a small 1-sentence description of each env purpose.

Definitely not a blocker now, but it would be nice to get rid of config duplication, especially in deployment scripts.

itegulov (dismissed):

kmaus-near (dismissed):

kmaus-near (dismissed):

kmaus-near (approved):

kmaus-near (approved):

PR #325, opened 2023-10-18
new_acc load tests

volovyks:


load-tests/Cargo.toml
```diff
@@ -8,3 +8,13 @@ goose = "^0.17"
...
 reqwest = "^0.11"
+jsonwebtoken = "8.3.0"
+near-primitives = "0.17.0"
+serde = "1.0.130"
+serde_json = "1.0.68"
+chrono = "0.4.19"
+near-crypto = "0.17.0"
+mpc-recovery = { path = "../mpc-recovery" }
+rand = "0.8.4"
+log = "0.4"
```
itegulov, line 19: Not a huge deal, but we don't use `log` in other crates, we use `tracing` instead

load-tests/src/main.rs
```diff
@@ -1,25 +1,160 @@
...
+        .register_scenario(
+            scenario!("registrationAndSign")
+                .register_transaction(transaction!(prepare_user_credentials).set_sequence(1))
+                .register_transaction(transaction!(claim_oidc).set_sequence(2))
+                .register_transaction(transaction!(new_account).set_sequence(3))
+                .register_transaction(transaction!(user_credentials).set_sequence(4))
+                .register_transaction(
+                    transaction!(sign)
+                        .set_sequence(5)
+                        .set_weight(1000) // In this scenario we are mostly testing /sign functionality
```
itegulov, line 55: What does `set_weight` actually do here? I would assume that these 5 actions above need to be executed in sequence, so why would we prioritize the last step over the others?

itegulov (approved):
Ran the load tests locally and confirm that they work. Awesome work!

load-tests/src/main.rs
volovyks: In the terminal, you are passing a number of users (N) and a scenario (`registrationAndSign` in this case). Then each user executes transactions in sequence (`check .set_sequence(k)`). The scenario starts from the beginning after the user ends the previous one.

In these particular load tests we are trying to make as many sign requests as possible. Without `.set_weight(1000)` sign will be executed once per registration and registration takes several seconds.

.gitignore
```diff
@@ -1,3 +1,4 @@
 /target
 .direnv
 .DS_Store
+tmp
```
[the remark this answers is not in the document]
volovyks: Nice! I will invest time into this if the current relayer setup remains permanent.

load-tests/Cargo.toml
volovyks: Good point. Done.

itegulov (approved):

PR #334, opened 2023-10-23
remove oidc providers check from signing nodes

volovyks:
Details are in the issue.


mpc-recovery/src/oauth.rs
```diff
@@ -57,8 +57,17 @@ fn validate_jwt(
...
+    match oidc_providers {
+        Some(oidc_providers) => {
+            if !oidc_providers.contains(issuer, audience) {
+                anyhow::bail!("UnauthorizedTokenIssuerOrAudience: iss={issuer}, aud={audience}");
+            }
+        }
+        None => {
+            // If no OIDC providers are specified in the allowlist, we allow any issuer and audience.
+            // Should be used in signing nodes only.
+            tracing::warn!("No OIDC providers found in the allowlist");
```
ChaoticTempest, line 69: this will produce a lot of noise in the logs, since this will be ran on every sign. We should have it at node initialization instead

ChaoticTempest (approved):

itegulov (dismissed):

volovyks:
After additional commits, I need new approvals. I assume this is something enforced by the Security team.

DavidM-D (approved):

PR #335, opened 2023-10-26
feat: implement Beaver triple generation

### LINKED ISSUE #328: Implement Beaver triple generation
Beaver triple generation is the most network-intensive part of the protocol. Each triple requires about 5 rounds of all-to-all ($`n^2`$) communication between participants. So it is imperative to make this efficient.

### Invariants
* Beaver triples should **never** be reused (even if the process failed mid through), so it is important to have mutually exclusive access to each of them.
* Leaking Beaver triples **does not** leak the private key share. Meaning it might(?) be okay to store them plaintext in a persistent storage.
* Beaver triples do not require the payload to be known yet, but for the sake of this ticket you can assume it is already present

### Protocol
A set of parties $`\mathcal{P}`$ of size $`N`$ make up the Multichain network. Any subset of $`\mathcal{P}`$ of size $`t`$ (called threshold) can reconstruct the private key. In our scheme signing is split into three phases:
* Triple generation
* Presigning
* Signing

Each of these phases can potentially use a different set of parties, and these sets can have different sizes. Furthermore, each phase can have a different threshold describing how large the party set for the next phase needs to be.

These sets, their sizes and thresholds should satisfy the following:

$$
\begin{matrix}
&\scriptsize{\text{Key-Gen}}
&&\scriptsize{\text{Triples}}
&&\scriptsize{\text{Presigning}}
&&\scriptsize{\text{Signing}}
\cr
&\mathcal{P} &\supseteq &\mathcal{P}_0 &\supseteq &\mathcal{P}_1 &\supseteq &\mathcal{P}_2\cr
&N &\geq &N_0 &\geq &N_1 & \geq &N_2\cr
&t &&t_0 &&t_1 &&t_2\cr
\end{matrix}
$$

Meaning that the subset can potentially shrink with every phase as long as the final subset is big enough to satisfy `t_2`. The relations around $`t`$, $`t_1`$, $`t_2`$ and $`t_2`$ are a bit complicated and are irrelevant for the context of this issue where we will consider them to be equal to each other and will just call it `t` from now on.

Furthermore, for the sake of simplicity I think it makes sense to always generate triplets with the full set of participants $`\mathcal{P}`$ and only narrow it to `t` at the very last phase.

Relevant API from cait-sith: https://github.com/LIT-Protocol/cait-sith/blob/main/src/triples/generation.rs#L451-L455
Do **NOT** use [this](https://github.com/LIT-Protocol/cait-sith/blob/main/src/triples/mod.rs#L66-L70) API outside of tests as it assumes a trustful party that will forget the values it generated. 

### Proposed implementation
A separate state machine similar to `NodeState` that regulates the current signing state. Something like:
* Starting(payload)
* GeneratingTriple1(payload, triple_protocol)
* GeneratingTriple2(payload, triple_1, triple_protocol)
* Presigning(payload, presign_protocol)
* Signing(payload, sign_protocol)
* Completed(payload, signature)

This ticket should only handle the first three phases.

This state machine will then be used inside of `Running` state under `NodeState`. Where in the future we can potentially have multiple signing state machines progressing at the same time.

You can presume that you will get payload through an external event.
TODO: link indexing ticket here once it exists

itegulov:
Fixes #328 

Introduces a separate entity `TripleManager` that tries to progress all ongoing triple generation protocols and saves their results. Triples are identified by a randomly generated id one the proposing node's side.


node/src/protocol/cryptography.rs
```diff
@@ -163,6 +163,23 @@ impl CryptographicProtocol for ResharingState {
...
+        mut self,
+        ctx: C,
+    ) -> Result<NodeState, CryptographicError> {
+        if self.triple_manager.potential_len() < 2 {
+            self.triple_manager.generate();
+        }
+        for (p, msg) in self.triple_manager.poke() {
+            let url = self.participants.get(&p).unwrap();
+            http_client::message(ctx.http_client(), url.clone(), MpcMessage::Triple(msg)).await?;
+        }
```
ChaoticTempest, line 178: is this correct that we should be doing all the messaging after the protocol completes? What if all nodes are waiting on messages and we'll be stuck in a deadlock?

node/src/protocol/triple.rs
```diff
@@ -0,0 +1,155 @@
...
+            Entry::Vacant(_) => None,
+            Entry::Occupied(entry) => Some(entry.remove()),
+        }
+    }
+
+    /// Ensures that the triple with the given id is either:
+    /// 1) Already generated in which case returns `None`, or
+    /// 2) Is currently being generated by `protocol` in which case returns `Some(protocol)`, or
+    /// 3) Has never been seen by the manager in which case start a new protocol and returns `Some(protocol)`
+    // TODO: What if the triple completed generation and is already spent?
```
ChaoticTempest, line 84: we probably need a cache just to see if the triple already got generated, or that triple never gets generated again

node/src/protocol/triple.rs
```diff
@@ -0,0 +1,155 @@
...
+use cait_sith::protocol::{Action, Participant};
+use cait_sith::triples::TripleGenerationOutput;
+use k256::Secp256k1;
+use std::collections::btree_map::Entry;
+use std::collections::BTreeMap;
+
+/// Unique number used to identify a specific ongoing triple generation protocol.
+/// Without `TripleId` it would be unclear where to route incoming cait-sith triple generation
+/// messages.
+pub type TripleId = u64;
```
ChaoticTempest, line 13: Would this id also be used to track the uniqueness of the triple used?

node/src/protocol/cryptography.rs
```diff
@@ -163,6 +163,23 @@ impl CryptographicProtocol for ResharingState {
...
 
+#[async_trait]
+impl CryptographicProtocol for RunningState {
+    async fn progress<C: CryptographicCtx + Send + Sync>(
+        mut self,
+        ctx: C,
+    ) -> Result<NodeState, CryptographicError> {
+        if self.triple_manager.potential_len() < 2 {
+            self.triple_manager.generate();
+        }
```
ChaoticTempest, line 174: so, we're generating more triples on the fly when we run low? Won't this be very computationally expensive or since we're just generating one, it will be fine? But wouldn't that still impose a good amount of latency with all the messaging the triple generation protocol requires?

node/src/protocol/triple.rs
volovyks: Triples are stored in RAM. It means that there is no guarantee that this cache was not wiped out by reboot. And it would be nice to avoid DB.

node/Cargo.toml
```diff
@@ -19,6 +19,7 @@ clap = { version = "4.2", features = ["derive", "env"] }
 hex = "0.4.3"
 k256 = { version = "0.13.1", features = ["sha256", "ecdsa", "serde"] }
 local-ip-address = "0.5.4"
+rand = "0.8"
```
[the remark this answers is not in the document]
itegulov: I don't think we need both versions anymore, but in the past there was some issue with one of the dependencies relying on rand being exactly 0.7

node/src/protocol/triple.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    /// Take an unspent triple by its id with no way to return it.
+    /// It is very important to NOT reuse the same triple twice for two different
+    /// protocols.
+    pub fn take(&mut self, id: TripleId) -> Option<TripleGenerationOutput<Secp256k1>> {
+        match self.triples.entry(id) {
+            Entry::Vacant(_) => None,
+            Entry::Occupied(entry) => Some(entry.remove()),
+        }
+    }
```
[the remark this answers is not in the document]
itegulov: lol yea just had a brainfart

node/src/web/mod.rs
```diff
@@ -124,3 +125,36 @@ async fn join(
...
+        participants: Vec<Participant>,
+        triple_count: usize,
+    },
+    NotRunning,
+}
+
+#[tracing::instrument(level = "debug", skip_all)]
+async fn state(Extension(state): Extension<Arc<AxumState>>) -> (StatusCode, Json<StateView>) {
+    tracing::debug!("fetching state");
+    let protocol_state = state.protocol_state.read().await;
```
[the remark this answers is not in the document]
itegulov: Can we have `Arc<RwLock<ProtocolState>>` in one place (where we write) and `Arc<ProtocolState>` in the other place where we only read tho?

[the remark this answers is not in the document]
itegulov: Just to clarify, it is shared with `MpcSignProtocol` inside `protocol/mod.rs`

node/src/protocol/cryptography.rs
itegulov: No, the protocol is interactive and occasionally you need to wait for other messages to arrive before you can progress. So ideally there should be a timeout and a restart mechanism but that hasn't been implemented yet.

node/src/protocol/triple.rs
itegulov: Yeah DB will come down the line, it is one of the tickets I have not scoped out yet

node/src/protocol/triple.rs
itegulov: Yes, it is not the end of the world if two of them happen to have the same id (although it shouldn't happen, picking two 64-bit numbers that are the same is very inprobable). It will just mean that the triple generation for this specific triple is going to fail and the nodes would have to give up on it.

node/src/protocol/cryptography.rs
itegulov: So this is just pretty much a placeholder. We need to implement `TripleStockpile` that operates on top of `TripleManager` and actively tries to initiate generation when it can (the ticket is in the epic, but hasn't been scoped out yet).

itegulov:
@ChaoticTempest sorry somehow I completely missed your comments! Addressed them in #348

PR #342, opened 2023-10-31
feat: implement presignature generation

### LINKED ISSUE #341: Implement presignature generation
### Description

A presignature is an interactive protocol with a prerequisite of two Beaver triples, but it can be computed offline. Requires 1 round of all-to-all ($`n^2`$) communication between participants.

#### Invariants
* Presignatures should never be reused (even if the process failed mid through), so it is important to have mutually exclusive access to each of them.
* Leaking presignature **leaks the private key share**, so it is very important to invest into a security layer for them
* Presignatures do not require the payload to be known yet

#### Challenges
* **Consensus on which triples to use.** Since nodes can start generating presignatures asynchronously they can cannibalize each other's ability to produce presignatures by using an overlapping set of triples. To overcome this we can mandate that every participant should only use triples proposed by them specifically.
* **Identification of presignatures.** We need to route networking messages to generation protocol, so we need some way to distinguish presignatures before they are actually generated. One way to do this is by indentifying them by a pair `(triple0, triple1)` which should work, but I propose to use a separately generated random presignature id to reuse it at a later stage for signatures.

#### For future
This issue is focused on the core functionality, persistence layer and stockpiling are left for the future work

itegulov:
Fixes #341


node/src/protocol/presignature.rs
```diff
@@ -0,0 +1,228 @@
...
+            },
+        );
+    }
+
+    /// Ensures that the presignature with the given id is either:
+    /// 1) Already generated in which case returns `None`, or
+    /// 2) Is currently being generated by `protocol` in which case returns `Some(protocol)`, or
+    /// 3) Has never been seen by the manager in which case start a new protocol and returns `Some(protocol)`, or
+    /// 4) Depends on triples (`triple0`/`triple1`) that are unknown to the node
+    // TODO: What if the presignature completed generation and is already spent?
```
volovyks, line 112: Also, can it happen that presignature generation is completed but it was not spent?

node/src/protocol/presignature.rs
```diff
@@ -0,0 +1,228 @@
...
+        &mut self,
+        id: PresignatureId,
+        triple0: TripleId,
+        triple1: TripleId,
+        triple_manager: &mut TripleManager,
+        public_key: &PublicKey,
+        private_share: &PrivateKeyShare,
+    ) -> Option<&mut PresignatureProtocol> {
+        if self.presignatures.contains_key(&id) {
+            None
```
volovyks, line 123: This part is a bit confusing. And I saw this pattern in other places.
Why we are not returning the presignature her? Is it because the signature generation part is not ready?

volovyks (approved):
We can discuss some of these questions at the meeting.

node/src/protocol/triple.rs
```diff
@@ -64,19 +80,47 @@ impl TripleManager {
...
         match self.triples.entry(id) {
             Entry::Vacant(_) => None,
             Entry::Occupied(entry) => Some(entry.remove()),
         }
     }
 
+    /// Take two random unspent triple generated by this node. Either takes both or none.
+    /// It is very important to NOT reuse the same triple twice for two different
+    /// protocols.
+    pub fn take_mine_twice(&mut self) -> Option<(Triple, Triple)> {
```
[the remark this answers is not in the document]
itegulov: Generating presignature requires two secret multiplications, so we need two triples. I implemented this is a single function to ensure that we take two triples atomically as there is no way to put them back once you have taken one out.

node/src/protocol/presignature.rs
```diff
@@ -0,0 +1,228 @@
...
+/// A completed presignature.
+pub struct Presignature {
+    pub id: u64,
+    pub output: PresignOutput<Secp256k1>,
+}
+
+/// An ongoing presignature generator.
+pub struct PresignatureGenerator {
+    pub protocol: PresignatureProtocol,
+    pub triple0: TripleId,
```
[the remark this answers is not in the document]
itegulov: Yes, as explained in the other thread

node/src/protocol/presignature.rs
itegulov: If the presignature is already generated, then we no longer have a valid `PresignatureProtocol` for it (i.e. protocol has been completed and destroyed)

node/src/protocol/presignature.rs
itegulov: Yes, that's the first case: `1) Already generated in which case returns None`

node/src/protocol/triple.rs
```diff
@@ -64,19 +80,47 @@ impl TripleManager {
...
+        tracing::info!(id0, id1, "trying to take two triples");
+        if self.triples.contains_key(&id0) && self.triples.contains_key(&id1) {
+            Some((
+                self.triples.remove(&id0).unwrap(),
+                self.triples.remove(&id1).unwrap(),
+            ))
+        } else {
+            tracing::warn!(id0, id1, "my triples are gone");
+            None
+        }
```
ChaoticTempest, line 121: This won't have those pesky unwraps:
```suggestion
        if let (Some(triple0), Some(triple1)) = (self.triples.remove(&id0), self.triples.remove(&id1)) {
            Some((triple0, triple1))
        } else {
            tracing::warn!(id0, id1, "my triples are gone");
            None
        }
```

node/src/protocol/triple.rs
```diff
@@ -2,25 +2,40 @@ use super::message::TripleMessage;
...
 /// complete some time in the future and a way to take an already generated triple.
 pub struct TripleManager {
     /// Completed unspent triples
-    // TODO: I put these into `BTreeMap` so we can potentially use the ordering for choosing which
-    // triple to use for consensus, but it might not be necessary.
-    triples: BTreeMap<TripleId, TripleGenerationOutput<Secp256k1>>,
+    triples: HashMap<TripleId, Triple>,
     /// Ongoing triple generation protocols
-    generators: BTreeMap<TripleId, TripleProtocol>,
+    generators: HashMap<TripleId, TripleGenerator>,
```
ChaoticTempest, line 36: so we know we no longer need the guarantees of BTreeMaps now?

node/src/protocol/presignature.rs
```diff
@@ -0,0 +1,228 @@
...
+                        &mut entry
+                            .insert(PresignatureGenerator {
+                                protocol,
+                                triple0: triple0.id,
+                                triple1: triple1.id,
+                            })
+                            .protocol,
+                    )
+                }
+                Entry::Occupied(entry) => Some(&mut entry.into_mut().protocol),
```
ChaoticTempest, line 168: Seems like we can just reuse the `generate` function instead of rewriting it here

ChaoticTempest (approved):
looks good to me besides a couple minor nits

node/src/protocol/triple.rs
ChaoticTempest: wait, nevermind, this isn't the same logic

itegulov: yeah I am still not sure if there is a way to denote this without unwraps

node/src/protocol/triple.rs
itegulov: Yeah turns out ordering is not necessary

node/src/protocol/presignature.rs
```diff
@@ -0,0 +1,239 @@
...
+            match self.generators.entry(id) {
+                Entry::Vacant(entry) => {
+                    tracing::info!(id, "joining protocol to generate a new presignature");
+                    let triple0 = match triple_manager.take(triple0) {
+                        Some(triple0) => triple0,
+                        None => {
+                            tracing::warn!(triple_id = triple0, "triple0 is missing, can't join");
+                            return Ok(None);
+                        }
+                    };
```
ChaoticTempest, line 150: nit: let-else could be used here
```suggestion
                    let Some(triple0) = triple_manager.take(triple0) else {
                        tracing::warn!(triple_id = triple0, "triple0 is missing, can't join");
                        return Ok(None);
                    };
```

ChaoticTempest (approved):

PR #345, opened 2023-11-01
feat: use Grafana OTLP collector

itegulov:
Do not merge yet, just trying to see if it works with a feature env

### DIFF
--- a/infra/terraform-dev.tfvars
+++ b/infra/terraform-dev.tfvars
@@ -20,5 +20,5 @@ signer_configs = [
   }
 ]
 jwt_signature_pk_url = "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com"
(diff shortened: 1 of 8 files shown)


infra/terraform-dev.tfvars
```diff
@@ -20,3 +20,5 @@ signer_configs = [
   }
 ]
 jwt_signature_pk_url = "https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com"
+otlp_endpoint        = "https://otel.dev.api.pagoda.co:443/v1/traces"
+opentelemetry_level  = "debug"
```
volovyks, line 24: Hm... should it be info? We are not very strict with levels though :/

volovyks (approved):

infra/terraform-dev.tfvars
itegulov: I was thinking to use info for mainnet/testnet and debug for dev just so we have more information there

Issue #343, opened 2023-10-31
All Docker image versions should be have fixed version


volovyks:
We should specify what version of Sandbox, Relayer, etc. we are using. That should have to make our testing env consistent on all machines and CI.

itegulov:
"Specify" as in use a specific version in the code instead of depending on latest or as in writing the version in README?

volovyks:
@itegulov first one, hardcode versions. I have faced inconsistencies several times.

ChaoticTempest:
serhii and I was running into issues with running the multichain tests. It turns out to be that we forgot to update to the latest snadbox. Tagging a specific docker version would probably help this out by erroring out what really went wrong.

Also, additionally, we should tag specific terraform versions too, just so that when we upgrade secrets or any other value, we don't interrupt someone else's PR

volovyks:
Now sandbox does not have real tags, only "latest". Here is the PR that should fix this: https://github.com/near/sandbox/pull/77

PR #347, opened 2023-11-02
Relayer hack deleted

volovyks:


volovyks:
It's super strange that tests passed. #343 seems even more important now.

volovyks:
Okay, the tests are passing with the new version of the relayer as well. I think we can merge this and continue discussions on the relayer side.

volovyks:
Okay, the tests are passing with the new version of the relayer as well. I think we can merge this and continue discussions on the relayer side.

itegulov (approved):

PR #348, opened 2023-11-09
fix: address Beaver triple generation comments

itegulov:
Addressed comments from #335


volovyks (approved):
I remember that we added another version of rand because of some outdated dep. But if the tests will pass - this issue I gone.

PR #351, opened 2023-11-11
feat: add tracecontext middleware for axum

### LINKED ISSUE #199: Attach a request id to response headers
In order to be able to trace errors thrown on the frontend, it's useful to have a request id header in the response that will be rendered in error toasts that the user sees. We can then query the backend logs based on the ids and trace the error more effectively.

caused by https://github.com/near/fast-auth-signer/issues/15

itegulov:
Fixes #199 

Spent hours on this and finally got it working. For some reason, OTLP middleware needs specific directives in its layer to be enabled to work... The commits are a mess, so just ignore them and review the end result.


itegulov:
You can test tracecontext behaviour by doing a request like this:
```
curl -X POST -H 'Content-type: application/json' -H 'traceparent: 00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-01' https://mpc-recovery-leader-dev-351-7tk2cmmtcq-ue.a.run.app/mpc_public_key -v -d "{}"
```

It should respond with a traceparent header matching the provided one. You can also leave the request header out in which case it will be randomly generated on the backend.

volovyks (approved):
Nice! Great work.
@esaminu we should have ids after the release.

PR #354, opened 2023-11-13
feat: secure p2p

ChaoticTempest:
This implements secure p2p communication between two nodes.

- ~Adds `/msg-private` endpoint for secure messaging on protocol `SendPrivate` actions.~ `msg` endpoint is now encrypted when sending messages
- State transitions are now transactional when progressing the protocol. A clone is made on the state each time, with the exception of triple managers and each protocols which have Arc/RwLock on top of them.


keys/src/hpke.rs
```diff
@@ -0,0 +1,156 @@
+use borsh::{self, BorshDeserialize, BorshSerialize};
```
itegulov, line 1: FYI I want Michel to take a look at this to make sure we are not misusing anything here, will let you know how it goes

itegulov (commented):
Haven't finished reviewing yet, but this is my thoughts so far

contract/src/lib.rs
```diff
@@ -5,6 +5,10 @@ use std::collections::{HashMap, HashSet};
 
 type ParticipantId = u32;
 
+pub mod hpke {
+    pub type PublicKey = [u8; 32];
```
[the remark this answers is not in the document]
ChaoticTempest: Yeah, it was a whole ordeal to make use of the `keys` crate to provide this one simple type. So I just hardcoded it for now since it wasn't worth the effort to put it inside the crate. There's probably some artifacts left from that

node/src/protocol/cryptography.rs
```diff
@@ -173,12 +195,29 @@ impl CryptographicProtocol for RunningState {
...
-        for (p, msg) in self.triple_manager.poke()? {
-            let url = self.participants.get(&p).unwrap();
-            http_client::message(ctx.http_client(), url.clone(), MpcMessage::Triple(msg)).await?;
+        for (is_public, p, msg) in self.triple_manager.poke()? {
+            let info = self
+                .participants
+                .get(&p)
+                .ok_or(CryptographicError::UnknownParticipant(p))?;
+            if is_public {
+                http_client::message(ctx.http_client(), info.url.clone(), MpcMessage::Triple(msg))
```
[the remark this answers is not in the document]
ChaoticTempest: yeah, probably better to just encrypt everything. The overhead of encrypting messages shouldn't be that big either

keys/src/hpke.rs
ChaoticTempest: sounds good

keys/src/hpke.rs
```diff
@@ -0,0 +1,156 @@
...
+    aead::{AeadTag, ChaCha20Poly1305},
+    kdf::HkdfSha384,
+    kem::X25519HkdfSha256,
+    OpModeR,
+};
+use serde::{Deserialize, Serialize};
+
+/// This can be used to customize the generated key. This will be used as a sort of
+/// versioning mechanism for the key.
+const INFO_ENTROPY: &[u8] = b"session-key-v1";
```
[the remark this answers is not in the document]
ChaoticTempest: this is just extra info for generating the derived key fro encryption. So it's more for contextual info. It's useful when we end up using the same key as encryption and signing, but we don't se it can just empty or static like this. I just arbitrarily made it into a versioning scheme

contract/src/lib.rs
```diff
@@ -22,6 +26,10 @@ pub struct ParticipantInfo {
     pub id: ParticipantId,
     pub account_id: AccountId,
     pub url: String,
+    /// The public key used for encrypting messages.
+    pub cipher_pk: hpke::PublicKey,
+    /// The public key used for verifying messages.
```
volovyks, line 31: I would rather have longer names here and in other places, like `msg_encryption_pk`, `msg_signature_pk`. People and other developers can confuse these with key shares.

keys/src/hpke.rs
volovyks: @itegulov  Are we good to go?

node/src/protocol/cryptography.rs
```diff
@@ -22,6 +25,21 @@ pub enum CryptographicError {
     CaitSithInitializationError(#[from] InitializationError),
     #[error("cait-sith protocol error: {0}")]
     CaitSithProtocolError(#[from] ProtocolError),
+    #[error("sync failed: {0}")]
+    SyncError(String),
+    #[error(transparent)]
+    DataConversion(#[from] serde_json::Error),
```
volovyks, line 31: nit: `<name>Error(`

itegulov:
Ok, I am merging this to avoid conflicts and we can address comments in a follow up PR

Issue #346, opened 2023-11-01
Index request transaction for multichain


itegulov:
### Description

Assuming that developers will not be able to request MPC signatures via a system call, we will have to somehow make MPC network aware of requests happening on-chain. One simple way to do this is just by indexing all transactions happening on `multichain.near` (placeholder name) and look for `sign(payload)` specifically (see the flow in https://github.com/near/mpc-recovery/issues/326 for full context).

There are two options for indexing in the NEAR ecosystem:
1. NEAR Indexer Framework
2. NEAR Lake Framework

See the comparison table from the [docs](https://docs.near.org/concepts/advanced/near-lake-framework):
![latest-screenshot](https://github.com/near/mpc-recovery/assets/7879134/3e19d854-7504-42c9-8f27-3df28f01790e)

As you can see, there are certain pros and cons which I would like to analyze before we commit to a specific solution. I can see three way how we can go about this:

#### (Option 1) Tightly coupled NEAR Indexer Framework

We use `near-indexer` crate directly in our code, making `mpc-recovery` a stateful service that embed nearcore inside.

Pros:
* Easy way to access indexer output using an already built Rust API
* Close to zero overhead
* Completely decentralized - every node runs its own built-in indexer

Cons:
* `mpc-recovery` will have to store hundreds of GBs of onchain data (assuming we only store data from the last ~24 hours) and potentially will require a few days to catch-up from scratch
* Complicated infrastructure setup (having to manage indexer's dependencies as a part of `mpc-recovery`)

#### (Option 2) Separate NEAR Indexer Framework microservice

We create a separate microservice similar to https://github.com/near-examples/indexer-tx-watcher-example that just indexes a specific contract and provides an API for new transactions there. Whether the API should be streaming, polling or callback is up to debate. Every node

Pros:
* Small overhead if done right
* Potentially decentralized - every node can run their own instance
* We can keep our current predicate of stateless `mpc-recovery` that can be run in Cloud Run, making complicated infra management concentrated in the new indexer service. Assuming nothing terrible happens to the service we will never have to wait multiple days for it to catch up. But even if we do, then other working nodes can just kick us out while this particular node is not operational.

Cons:
* Engineering work required to build this
* One more entity to manage and track metrics of

#### (Option 3) NEAR Lake Framework

We just stream data from S3 using an already built Rust client.

Pros:
* Almost zero effort required to set this up

Cons:
[...]

PR #358, opened 2023-11-14
feat: use local Lake Indexer in integration tests

itegulov:
Relates to #346 

Sets up local instance of NEAR Lake Indexer that dumps all of the data to LocalStack S3 which is then consumed by the multichain node


ChaoticTempest (dismissed):
LGTM: seems like there's a couple cleanups we can do later. That `containers.rs` file is getting pretty big...

ChaoticTempest (approved):

PR #361, opened 2023-11-15
Allow all actions in signing nodes

volovyks:
- Action type check deleted on the signing node level
- Account deletion prevented on the leader node level

### DIFF
--- a/mpc-recovery/src/error.rs
+++ b/mpc-recovery/src/error.rs
@@ -71,6 +71,8 @@ pub enum LeaderNodeError {
     RelayerError(#[from] RelayerError),
     #[error("recovery key can not be deleted: {0}")]
     RecoveryKeyCanNotBeDeleted(PublicKey),
+    #[error("action can not be performed, account deletion is not allowed")]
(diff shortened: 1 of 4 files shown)


volovyks:
@itegulov looks like test are broken in develop

mpc-recovery/src/error.rs
```diff
@@ -71,6 +71,8 @@ pub enum LeaderNodeError {
     RelayerError(#[from] RelayerError),
     #[error("recovery key can not be deleted: {0}")]
     RecoveryKeyCanNotBeDeleted(PublicKey),
+    #[error("action can not be performed, account deletion is not allowed")]
+    AccountDeletionIsNotAllowed,
```
ChaoticTempest, line 75: nit: `IsNotAllowed` seems a little funky to type out. How about `Unsupported` since it's not as rough to type out with fewer capitalizations and one letter shorter

ChaoticTempest (approved):

itegulov (approved):

PR #362, opened 2023-11-15
Update terraform-dev.yml

kmaus-near:
Added an additional directory traverse to fix docker build

### DIFF
--- a/.github/workflows/terraform-dev.yml
+++ b/.github/workflows/terraform-dev.yml
@@ -57,7 +57,7 @@ jobs:
       # Build Docker image.
       - name: Docker Image
         id: build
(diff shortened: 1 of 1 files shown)


itegulov (approved):

.github/workflows/terraform-dev.yml
```diff
@@ -57,7 +57,7 @@ jobs:
       # Build Docker image.
       - name: Docker Image
         id: build
-        run: docker build .. -t near/mpc-recovery
+        run: docker build ../.. -t near/mpc-recovery
```
itegulov, line 60: @kmaus-near Why do we need to build this image at all? Aren't we using Artifact Registry one in the next step?

PR #380, opened 2023-11-22
feat: implement signing

### LINKED ISSUE #385: Implement signature generation
### Description

Direct continuation of #328 and #341 (see both for context).

Given a presignature, message hash (see #346) along with some source of public entropy, we can generate a signature by following a quick interactive protocol.

Public entropy is used to select a subset of size $t$ and a specific participant in that subset that will select a presignature and start the generation protocol. This participant is also responsible for publishing the response on chain.

Note that a node can only propose signature protocol using a presignatures it "owns", so if a node is starved for triples/presignatures it might take some time for it to actually propose. We might want to add a mechanism where a node can delegate its right to generate a signature to someone else.

itegulov:
Apologies that this code is pretty hairy and that PR is touching a million different things, but I had to fix many little bugs to make this actually work.

I tried to serialize the resulting signature into something more conventional for NEAR's ecosystem, but spent 3 hours going through different tags, SEC1 variations and compression without writing anything that works. So I am leaving that for a follow-up PR.


volovyks:
@itegulov do yo want to merge it with the failing test and then fix it?

integration-tests/tests/multichain/mod.rs
```diff
@@ -49,3 +56,65 @@ async fn test_triples_and_presignatures() -> anyhow::Result<()> {
...
+            let worker = &ctx.nodes.ctx().worker;
+            let (account_id, secret_key) = worker.dev_generate().await;
+            worker
+                .create_tla(account_id.clone(), secret_key.clone())
+                .await?
+                .into_result()?;
+            let payload: [u8; 32] = rand::thread_rng().gen();
+            let outcome = ctx
+                .rpc_client
+                .send_tx(
```
volovyks, line 77: Since the protocol feature is yet to be implemented, what is our plan here (for this PR and first testnet release)?

volovyks (dismissed):

integration-tests/Cargo.toml
```diff
@@ -8,23 +8,22 @@ publish = false
 aes-gcm = "0.10"
 anyhow = { version = "1.0", features = ["backtrace"] }
 async-process = "1"
+aws-config = "0.54.0"
```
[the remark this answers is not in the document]
itegulov: Yes, needed for Lake Framework configuration

integration-tests/tests/multichain/mod.rs
itegulov: We are just making a separate unrelated transaction with the response

volovyks (approved):

PR #395, opened 2023-11-29
fix: Added queue for sending encrypted messages

ChaoticTempest:


itegulov (commented):
Hmm, I might be misunderstanding the code, but wouldn't it just crash after the first failed send?

ChaoticTempest:
@itegulov in the top level `state.advance` and `state.progress`, we have the error handling such that it would log the error. When the message fails to send in `MessageQueue`, it won't pop it, but keep it for next time, so once the protocol loop wraps around to the sending portion again, that's where it would retry

ChaoticTempest:
previous comment wasn't entirely correct. If failed in the top level section, the state won't be saved, so the failures would indeed abort. Recent commit directly handles the failed sending of encrypted messages at the call site instead

itegulov (approved):
LGTM

PR #396, opened 2023-11-29
feat: add persistent secret storage

itegulov:
Hacked up something to store the private key share in GCP Secret Manager. Not tested yet


itegulov:
Just deployed this branch to multichain0.testnet and it seems to be working fine.

Example:
sign - https://testnet.nearblocks.io/txns/6CFxjd7cxG3khcFgqj46yFETxPAG3ssDV3FYH4q4k6Ei
respond - https://testnet.nearblocks.io/txns/434d5XRRFMmp1okUXE3mMRiFpyws1EGJP2BbcLrLTdW3

integration-tests/Cargo.toml
```diff
@@ -32,7 +32,7 @@ serde_json = "1"
 testcontainers = { version = "0.15", features = ["experimental"] }
 tokio = { version = "1.28", features = ["full"] }
 tracing-subscriber = { version = "0.3", features = ["env-filter"] }
-near-workspaces = "0.8.0"
+near-workspaces = { git = "https://github.com/near/near-workspaces-rs.git", branch = "main" }
```
ppca, line 35: why?

contract/src/lib.rs
```diff
@@ -64,6 +64,7 @@ pub struct ResharingContractState {
 
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug)]
 pub enum ProtocolContractState {
+    NotInitialized,
```
volovyks, line 67: Why do we need one more state?

volovyks, line 67: Why do we need one more status?

volovyks (approved):
LGTM

node/src/protocol/cryptography.rs
```diff
@@ -118,6 +122,13 @@ impl CryptographicProtocol for GeneratingState {
                         public_key = hex::encode(r.public_key.to_bytes()),
                         "successfully completed key generation"
                     );
+                    ctx.secret_storage()
```
ppca, line 125: So if this storage attempt fails, the progress() will fail even tho it actually completed key generation. Is there a problem with that?
Also want to check the purpose of this change? Is it to have the nodes generate keys and upload themselves instead of us giving them keys?

node/src/storage.rs
```diff
@@ -0,0 +1,165 @@
...
+            ))
+            .doit()
+            .await?;
+        match response.payload {
+            // GCP does not allow to upload empty secrets, so we reserve 1-byte values as a
+            // placeholder for empty secrets.
+            Some(SecretPayload {
+                data: Some(data), ..
+            }) if data.len() > 1 => Ok(Some(serde_json::from_slice(&data)?)),
+            _ => {
```
ppca, line 117: if it's missing and return Ok(None), seems like the node will be started with None as the PersistentNodeData. Is that acceptable?

integration-tests/Cargo.toml
itegulov: workspaces 0.8.0/0.9.0 depends on NEAR SDK 4.x which is no longer compilable due to a yanked dependency. See https://github.com/near/near-sdk-rs/issues/1119.

So we have to use the unreleased version of workspaces for now

node/src/protocol/cryptography.rs
itegulov: Not ideal for sure, but I would say storage being unavailable is a fatal error. If we can't store the private key share, then we wouldn't be able to recover it after reboot.


> Also want to check the purpose of this change? Is it to have the nodes generate keys and upload themselves instead of us giving them keys?

So the idea is to be able to restart the node without loosing crucial information - current epoch, pre-agreed root public key and private key share specific for this node.

node/src/storage.rs
itegulov: Yeah, it just means we don't have any key share saved and we presume the node just doesn't have one yet (has to either generate one in collaboration with other participants, or try to join an existing set via resharing).

contract/src/lib.rs
itegulov: It's a stop-gap for now, otherwise it's very hard to deploy multichain to GCP.

itegulov: I will document this behaviour

infra/modules/multichain/main.tf
```diff
@@ -101,6 +101,14 @@ resource "google_cloud_run_v2_service" "node" {
         name  = "AWS_DEFAULT_REGION"
         value = var.indexer_options.s3_region
       }
+      env {
+        name  = "MPC_RECOVERY_GCP_PROJECT_ID"
+        value = var.project
+      }
+      env {
+        name  = "MPC_RECOVERY_SK_SHARE_SECRET_ID"
```
ChaoticTempest, line 109: should we rename these to be multichain instead of mpc-recovery?

contract/src/lib.rs
```diff
@@ -296,4 +296,15 @@ impl MpcContract {
 
     #[allow(unused_variables)]
     pub fn respond(&mut self, receipt_id: [u8; 32], big_r: String, s: String) {}
+
+    #[private]
+    #[init(ignore_state)]
+    pub fn clean(keys: Vec<near_sdk::json_types::Base64VecU8>) -> Self {
+        for key in keys.iter() {
+            env::storage_remove(&key.0);
```
ChaoticTempest, line 304: where is clean being called from? I wonder if we should return an error just in case `storage_remove` couldn't find the respective key

node/src/protocol/mod.rs
```diff
@@ -86,9 +88,13 @@ impl ConsensusCtx for &Ctx {
...
         &self.sign_sk
     }
+
+    fn secret_storage(&self) -> &SecretNodeStorageBox {
+        &self.secret_storage
+    }
 }
 
-impl CryptographicCtx for &Ctx {
+impl CryptographicCtx for &mut Ctx {
```
ChaoticTempest, line 97: nothing to change here, but now that I think about it, why weren't we just passing around a concrete `Ctx` object instead of requiring specific `*Ctx` interfaces? Doesn't seem like we're getting much out of having traits for this purpose but maybe I'm missing something

infra/modules/multichain/main.tf
itegulov: Mmm, not sure, maybe? I don't have a strong opinion so I just kept it as is for now

contract/src/lib.rs
itegulov: Just called externally via NEAR CLI, like in [this](https://gist.github.com/ilyar/19bdc04d1aa09ae0fc84eb4297df1a1d) guide

node/src/protocol/mod.rs
itegulov: The initial idea was to be able to "mock" specific methods that are necessary for that component without having to instantiate other stuff. Not sure if that makes sense anymore as most of the methods are used everywhere

ChaoticTempest: yeah, I kept having to add new interface members to each of the traits, so probably need to clean this up later. made a ticket: https://github.com/near/mpc-recovery/issues/406

infra/modules/multichain/main.tf
ChaoticTempest: Made a ticket for this one too: https://github.com/near/mpc-recovery/issues/407

node/src/protocol/cryptography.rs
ppca: `If we can't store the private key share, then we wouldn't be able to recover it after reboot.` -- Agree.
so when this progress() fails what happens next? Will progress() be tried again to generate and store the key share?

itegulov: Hmm, that's a good point. It might just make sense to panic here instead to signify that this is a fatal error. Otherwise progress will try to finish the protocol and save the result indefinitely.

PR #397, opened 2023-11-29
removed signer node ILBs

kmaus-near:


itegulov (dismissed):
Should we also delete the prod and testnet ones?

kmaus-near:
Good point, removed!

itegulov (approved):

PR #399, opened 2023-12-12
fix: decrypt message during joining phase

itegulov:
There is a race condition during resharing where a message can arrive before a joining realizes the contract has moved to `Resharing` phase. With this fix it will at least be able to decrypt the message and store it in the queue.


itegulov:
This should fix flaky tests that we are experiencing right now

volovyks (approved):

PR #400, opened 2023-12-13
Determine Participant Id automatically

volovyks:
- Now ParticipantId is determined by a position in the BTreeMap of Participants
- Since ParticipantId is not hard coded, but determined on the flight in an async way, some refactoring is done in consensus and other layers. For example - the consensus layer does not have the `me()` function because the participant id is not available at that moment.
- ParticipantId (me) is removed from Ctx, since it's not permanent. Replaced by AccountId.
- Nodes vote and communicate with the contract using AccountIds, for communication between nodes we are still using ParticipantIds.
- Changed how we work with context and node state. Both of them are needed to determine Node ParticipantId
- Used structs to represent Participants, Candidates, Votes, etc. in contract and node/contract
- Nodes on the contract layer are grouped by AccountId. node/contract layer is grouping them by ParticipantId


contract/src/primitives.rs
```diff
@@ -0,0 +1,197 @@
+use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
+use near_sdk::serde::{Deserialize, Serialize};
+use near_sdk::{AccountId, PublicKey};
+use std::collections::{BTreeMap, HashSet};
+
+pub mod hpke {
```
DavidM-D, line 6: What is `hpke`?

contract/src/primitives.rs
```diff
@@ -0,0 +1,197 @@
...
+    Hash,
+    PartialEq,
+    Eq,
+    PartialOrd,
+    Ord,
+    Debug,
+)]
+pub struct CandidateInfo {
+    pub account_id: AccountId,
+    pub url: String,
```
DavidM-D, line 58: If this is a URL, could it be a parsed one?

DavidM-D:
I might need a call to work out everything that's going on here. What issues does this PR close?

contract/src/primitives.rs
volovyks: It's used to determine the public key type used for encrypting messages.
Crate: https://docs.rs/crate/hpke/latest
(it is not added by this PR)

contract/src/primitives.rs
volovyks: I was avoiding it because such a check can be heavy for a contract. But it may be a good idea.

node/src/protocol/consensus.rs
```diff
@@ -310,6 +322,12 @@ impl ConsensusProtocol for WaitingForConsensusState {
                     }
                     let participants_vec: Vec<Participant> =
                         self.participants.keys().cloned().collect();
+
+                    let me = contract_state
+                        .participants
+                        .find_participant(ctx.my_account_id())
+                        .unwrap();
```
itegulov, line 329: This is going to be a brittle pattern going forward. `me()` does not exist in all contexts so we should strive to make this type safe and ensure we can only use it in the contexts where it exists. One potential solution to this is https://github.com/near/mpc-recovery/issues/416

itegulov (approved):
I had a session with @volovyks where he walked me through this PR and I think everything here makes sense to me. Except for one point, which is important but can be handled in the future.

PR #411, opened 2024-01-11
create terraform for multichain testnet and prod

ppca:
Mostly copy pasted from multichain-dev


itegulov (approved):
Looks good to me, but let's hold off on merging until we test it

PR #412, opened 2024-01-15
Randomly assign triples to a node after creation

### LINKED ISSUE #384: Use fair triple distribution mechanism
### Description

Currently we "assign" (we might want a better word for this?) a triple to the node that proposed it. It makes more sense to assign it to a random node based on the resulting `C` value of the triple. This way we can ensure that one fast node won't own all of the triples in the system.

DavidM-D:
Closes near/mpc#384


node/src/protocol/triple.rs
```diff
@@ -253,3 +258,131 @@ impl TripleManager {
...
+
+    // TODO: This test currently takes 22 seconds on my machine, which is much slower than it should be
+    // Improve this before we make more similar tests
+    #[test]
+    fn happy_triple_generation() {
+        let mut tm = TestManagers::new(5);
+
+        const M: usize = 2;
+        const N: usize = M + 3;
+        // Generate 23 triples
```
itegulov, line 328:
```suggestion
        // Generate 5 triples
```

itegulov (dismissed):
LGTM, but yeah we should look into why there are so many duplicate(?) messages being generated

DavidM-D:
Created an issue to understand why so many messages are getting sent at https://github.com/near/mpc-recovery/issues/415

itegulov (approved):

volovyks (commented):
@DavidM-D Can the node generate a triple, check if its entropy (big_c) will allow it to own it, and only then proceed?

itegulov:
@volovyks what do you mean by "then proceed"? You only get to know `big_c` by the point when the generation is fully completed. I suppose a node can decide not to store it's part of the triple, but it would be a purely malicious DOS action as there is no benefit to do that aside from saving 64 bytes in your database.

PR #413, opened 2024-01-16
feat: Added stockpiling beaver triples at the start

ChaoticTempest:
This adds stockpiling of beaver triples after generation or resharing phase. The triples are automatically cleaned after resharing since they are dropped and are not passed onto to the next state. The stockpiling uses somewhat of a naive logic to calculate how many triples to stockpile based on node size. There's a default max of 22500 messages that can be sent by all nodes. This is ideal for about 15 nodes being online, which yields 100 triples on each node. Less nodes will also cap out at 100 triples stockpiled at most. More nodes will have diminishing amount of stockpiling due to how long it will take. For example, 30 nodes will have 25 triples being stockpiled.

With this PR, `test_signature` stockpiling 100 triples with 3 nodes takes about 170s in total to complete the test, on an M1. Larger amounts of stockpiling takes up more time and leads to the test failing the deadline, which means we should either fix that test or ideally create separate states for this stockpiling phase, as to signal that it's not really in the running state yet in the test.


node/src/protocol/triple.rs
```diff
@@ -100,6 +106,15 @@ impl TripleManager {
         Ok(())
     }
 
+    pub fn generate_pile_by_bandwidth(&mut self, nodes: usize) -> Result<(), InitializationError> {
+        let pile = calc_optimal_pile(DEFAULT_MAX_MESSAGES, nodes);
+        tracing::info!(nodes, pile, "generating pile of triples");
+        for _ in 0..pile.min(DEFAULT_MAX_PILE) {
```
DavidM-D, line 112: Is this going to lock up the node or is this running on it's own thread?

ChaoticTempest: @itegulov correct me if I'm wrong, but it shouldn't lock up the node, as the triple protocol will just yield back a bunch of messages to be sent once poked, so the crypto/consensus loop should keep on continuing. As far as I saw, probably the only case that might've been bad is if it wants to generate a certain triple with a TripleId but the stockpile of triple has priority and the node needs to wait longer to generate the signature in that case

node/src/protocol/triple.rs
```diff
@@ -9,6 +9,12 @@ use std::collections::hash_map::Entry;
 use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
 
+/// Maximum number of messsages that should be relayed when generating a stockpile of triples.
+pub const DEFAULT_MAX_MESSAGES: usize = 22500;
+
+/// The pile of triples that should be generated by each node should not exceed this number.
```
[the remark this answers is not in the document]
ChaoticTempest: Was lazy in adding that, but will add it later if everything else looks good

node/src/protocol/triple.rs
```diff
@@ -9,6 +9,12 @@ use std::collections::hash_map::Entry;
 use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
 
+/// Maximum number of messsages that should be relayed when generating a stockpile of triples.
+pub const DEFAULT_MAX_MESSAGES: usize = 22500;
```
volovyks, line 13: Is this number determined empirically?

node/src/protocol/triple.rs
```diff
@@ -9,6 +9,12 @@ use std::collections::hash_map::Entry;
...
+
+/// The pile of triples that should be generated by each node should not exceed this number.
+pub const DEFAULT_MAX_PILE: usize = 100;
```
volovyks, line 16: Probably we will need this in our configs as well.

node/src/protocol/triple.rs
ChaoticTempest: I just chose 100 because I remember Michel saying that number was a good number for stockpiling but don't exactly remember where anymore. @itegulov would you happen to remember?

node/src/protocol/consensus.rs
```diff
@@ -310,19 +323,33 @@ impl ConsensusProtocol for WaitingForConsensusState {
...
+                    );
+                    // Start stockpiling triples in the background. This will wait until crypto loop to generate.
+                    if let Err(err) =
+                        triple_manager.generate_pile_by_bandwidth(participants_vec.len())
+                    {
+                        tracing::warn!(
+                            ?err,
+                            len = triple_manager.len(),
+                            potential_len = triple_manager.potential_len(),
+                            "failed to stockpile triples on startup: not fatal, but will slow down signing"
```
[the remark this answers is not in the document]
ChaoticTempest: I'll include the state of the contract so we should be able to distinguish the two

integration-tests/src/multichain/mod.rs
```diff
@@ -11,6 +11,21 @@ use std::collections::HashMap;
...
+pub struct MultichainConfig {
+    pub nodes: usize,
+    pub triple_stockpile: usize,
+}
+
+impl Default for MultichainConfig {
+    fn default() -> Self {
+        Self {
+            nodes: 3,
+            triple_stockpile: 10,
```
volovyks, line 24: Should we use that const here?

volovyks (dismissed):

ChaoticTempest:
I'ma hold off merging this PR until we have persistent triples. Stockpiling triples will be quite a huge time burn and a waste if the nodes restart. Also looking into making the triple generation process more efficient, so this PR can come in much later when things are more stable

ChaoticTempest:
So, I went back to having just passive stockpiling. This way, the logic to accrue triples is simple where we can have a minimum amount of triples owned by each node, and a maximum total amount among all nodes. The actual triple itself should be generated in a separate thread eventually, but for now, we have these configs to configure this logic

@ppca @volovyks

node/src/protocol/triple.rs
```diff
@@ -15,6 +15,12 @@ use std::collections::hash_map::Entry;
 use std::collections::{HashMap, VecDeque};
 use std::time::Instant;
 
+/// The minimum amount of triples that each node needs to own.
+pub const DEFAULT_MIN_TRIPLES: usize = 10;
```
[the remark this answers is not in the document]
ChaoticTempest: whoops these weren't supposed to have made it in. I'll delete in the next PR since I don't want to dismiss your approval here

node/src/test_utils.rs
```diff
@@ -15,6 +16,13 @@ use std::collections::VecDeque;
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
+// Constants to be used for testing.
+const STARTING_EPOCH: u64 = 0;
+const DEFAULT_TRIPLE_CONFIG: TripleConfig = TripleConfig {
+    min_triples: 2,
+    max_triples: 10,
```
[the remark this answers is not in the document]
ChaoticTempest: it should still work. It's just that the number of messages dramatically increases but we can later limit this with the amount of concurrent generation that can happen per node. We have the max_triples which is the total amount happening throughout all nodes right now

PR #417, opened 2024-01-16
chore: cleanup protocol types

ChaoticTempest:
The `{Triple, Presignature, Signature}Protocol` seems to not need to be behind a `Arc<RwLock<_>>` since they're only ever written to and never read and we always have a mutable reference in the crypto loop anyways. This also simplifies the code a bit

### DIFF
--- a/node/src/protocol/triple.rs
+++ b/node/src/protocol/triple.rs
@@ -9,7 +9,6 @@ use k256::elliptic_curve::group::GroupEncoding;
 use k256::Secp256k1;
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, VecDeque};
-use std::sync::Arc;
 
 /// Unique number used to identify a specific ongoing triple generation protocol.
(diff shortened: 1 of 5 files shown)


itegulov (approved):

node/src/protocol/triple.rs
```diff
@@ -160,19 +155,8 @@ impl TripleManager {
...
-                    Err(err) => {
-                        tracing::error!(
-                            ?err,
-                            "failed to acquire lock on triple generation protocol"
-                        );
-                        break false;
-                    }
-                };
-
                 let action = match protocol.poke() {
```
ppca, line 160: Isn't this a read of the protocol?

ChaoticTempest: Nope, the read of the protocol comes from explicit `read` calls since it was behind a lock

ppca: oh i see. protocol.read()

PR #421, opened 2024-01-17
chore: Expose root public key in mpc contract

ChaoticTempest:


DavidM-D (approved):
LGTM

ppca:
who do we expect would typically use this function?

itegulov:
Hmm, I don't think this was strictly necessary as `state()` was already exposing public key

itegulov:
@ppca 
> who do we expect would typically use this function?

Partners are asking how can they get to know what public key is going to be used to sign the payloads they are going to send into the system. In order to do that, they need to get the root public key from the contract and then run KDF on it to get the derived key they are interested in.

ChaoticTempest:
Ahh, I forgot about that `state` method, but I don't think it's bad to have since users would likely have to go directly into the state json and directly get the public key, meanwhile this will return a PublicKey or null, so it's pretty easy to use

PR #422, opened 2024-01-17
differentiate the logging messages 

ppca:
in node/src/protocol/consensus.rs and node/src/protocol/cryptography.rs
I basically added NodeState(ProtocolState) to the start of each message.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -17,7 +17,7 @@ dependencies = [
  "actix-macros",
  "actix-rt",
  "actix_derive",
- "bitflags 2.4.1",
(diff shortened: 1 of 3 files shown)


ChaoticTempest (approved):
LGTM

volovyks:
It's better to use spans: https://docs.rs/tracing/latest/tracing/
Can be a bit heavier though.

ppca:
> It's better to use spans: https://docs.rs/tracing/latest/tracing/ Can be a bit heavier though.

We can do this as a followup, or better enforce a consistent tracing style in our code by first changing all logging to the desired style.

PR #429, opened 2024-01-19
Use Participants only in a context where they actually exist

volovyks:
ParticipantId should not be available in the `Initializing` context.


ppca:
ParticipantId should not be available in the Initializing context. ---
what's the motivation for this?

volovyks:
@ppca now on both contract and node sides we have 4 states:
- NotInitialized,
- Initializing(InitializingContractState)
- Running(RunningContractState)
- Resharing(ResharingContractState)

ppca (dismissed):
LGTM. I think it'd be useful to add in PR title indicating contract change involved, so I'd know to redeploy contract as well.

volovyks:
@ppca agree
Do you think we should automate contract deployment? At least for dev env.

ppca (approved):

PR #440, opened 2024-01-27
chore: cleanup testing

ChaoticTempest:
This simplifies our testing layout a bit. This moves signature related generation to the `actions` module so that we have a clear view when writing integrations testing.


integration-tests/tests/multichain/mod.rs
```diff
@@ -68,89 +48,11 @@ async fn test_triples_and_presignatures() -> anyhow::Result<()> {
...
-
-            assert!(signature.verify(
-                &kdf::derive_key(public_key, epsilon),
-                &Scalar::from_bytes(&payload),
-            ));
-
-            Ok(())
+            wait_for::has_at_least_triples(&ctx, 2).await?;
+            wait_for::has_at_least_presignatures(&ctx, 2).await?;
+            actions::single_signature_production(&ctx, &state_0).await
```
volovyks, line 55: It feels too much for me. This top-level approach may not give us enough flexibility. I would rather move the code from this particular function to the test itself. We may add such an abstraction if we have a few more tests.

volovyks (approved):
LGTM

integration-tests/tests/multichain/mod.rs
ChaoticTempest: we can always degranularize the individual parts of the function. I already have a need for just running a single signature generation multiple times, so feels like this is a very common case when we want to test things on top of signature production

ChaoticTempest:
@volovyks can I get a reapproval

volovyks (approved):

PR #443, opened 2024-01-31
Contract logs

volovyks:


volovyks:
@ChaoticTempest @ppca any objection here?

node/src/indexer.rs
```diff
@@ -91,7 +91,7 @@ async fn handle_block(
                             tracing::warn!("`sign` did not produce entropy");
                             continue;
                         }
-                        let Ok(entropy) = serde_json::from_str::<'_, [u8; 32]>(&receipt.logs()[0])
+                        let Ok(entropy) = serde_json::from_str::<'_, [u8; 32]>(&receipt.logs()[1])
```
ChaoticTempest, line 94: hmm, why are we taking the second log entry instead of the first? What is the difference and can you show an example?

contract/src/lib.rs
```diff
@@ -277,14 +317,16 @@ impl MpcContract {
         payload: [u8; 32],
         depth: usize,
     ) -> PromiseOrValue<(String, String)> {
+        log!("sign_helper: payload={:?} depth={}", payload, depth);
```
ChaoticTempest, line 320: don't think this log is necessary since we already have one later with "signature not ready yet" along with the depth. The panic won't give us any logs anyways

node/src/indexer.rs
volovyks: I've added `"sign: signer={}, payload={:?} path={:?}"` log. Now entropy is the second log.

node/src/indexer.rs
```diff
@@ -91,7 +91,7 @@ async fn handle_block(
                             tracing::warn!("`sign` did not produce entropy");
                             continue;
```
ppca, line 92: should this check len() > 1 instead of empty?

ppca (approved):

ChaoticTempest (approved):

node/src/indexer.rs
volovyks: not sure what you mean here. can you please elaborate?

PR #444, opened 2024-01-31
persist triples

ppca:
Included in this PR:
1) when a node starts, it will load the triples from datastore and save it to the triples it has, and triple managers won't be changed. The node start won't crash if such load did not succeed, it will just proceed with empty set of triples.
2) when the node generates a triple or remove a triple, the change will be persisted to datastore, but failure to persist won't crash the triple generation/removal function.
3) refactored the GCP logic to instantiate secret storage and triple storage with one GcpService object and share parameters.
4) integration test now runs with a secret being saved in memory and triples saved to datastore emulator.

About tests: 
1) I've written tests that talk to our gcp datastore to load triples, insert and delete triples, they went thru. But left them out from PR since they relied on certain triples being present in the datastore and also the machine running the integration test must have our gcp authentification. 
2) I wanted to write in integration test to verify the datastore has persisted triples but not yet succeeded. I've put the err message in comment in the integration test, if folks have ideas let me know.
3) deployed my branch a set of 3 dev nodes, kill one of them, bring it back up to test the functionality. Seems to be working fine. Node here: https://console.cloud.google.com/run/detail/us-east1/multichain-dev-xiangyi-2/logs?project=pagoda-discovery-platform-dev

About encryption:
will do a followup for encryption after this one merges.


integration-tests/src/multichain/mod.rs
```diff
@@ -104,6 +114,17 @@ pub async fn setup(docker_client: &DockerClient) -> anyhow::Result<Context<'_>>
         .await?;
     tracing::info!(contract_id = %mpc_contract.id(), "deployed mpc contract");
 
+    let gcp_project_id = "multichain-integration";
+    let datastore =
+        crate::env::containers::Datastore::run(docker_client, docker_network, gcp_project_id)
+            .await?;
+
+    let storage_options = mpc_recovery_node::storage::Options {
+        gcp_project_id: Some("multichain-integration".to_string()),
```
volovyks, line 123: Probably we want to reuse the variable defined above.

integration-tests/tests/multichain/mod.rs
```diff
@@ -38,6 +38,15 @@ async fn test_triples_and_presignatures() -> anyhow::Result<()> {
             assert_eq!(state_0.participants.len(), 3);
             wait_for::has_at_least_triples(&ctx, 2).await?;
             wait_for::has_at_least_presignatures(&ctx, 2).await?;
+            // TODO: add test that checks #triples in datastore
```
volovyks, line 41: I would rather create a separate test. In that test, I would:
- Start 3 nodes
- Wait for a generation of N triples
- Stop all 3 nodes
- Start one of them and check that it has N triples (it will not be able to generate new triples, since other nodes are down)

node/src/gcp/value.rs
```diff
@@ -0,0 +1,383 @@
+use google_datastore1::api::{ArrayValue, Entity, Key, LatLng};
+use std::collections::HashMap;
+
+use super::error::ConvertError;
+
+#[derive(Debug, Clone)]
+pub enum Value {
+    BooleanValue(bool),
```
volovyks, line 8: Is it hard to create a place for common code between FastAuth and Multichain codebases? Code duplication is never a good idea. But I understand, we do not want to touch the old codebase.

volovyks (approved):
LGTM, all the comments are optional or can be addressed in a separate PR. 

Overall the only thing that concerns me with triple persistence is more edge cases, stale triples, etc. We should design it in a way that will not introduce new challenges for us and just help to reduce the number of triples that we need to generate after reboot.

- Do we have a mechanism to delete stale triples? Otherwise, our DB can grow indefinitely.
- I assume that if some of the triples are not in sync between all the nodes, we will drop this particular protocol, choose the other 2 triples, and try again. Is that so @ChaoticTempest ? It should allow us to overcome read/write failures to DB.

Integration tests:
- Separate tests for DB are not a priority, let's not waste time on this for now
- A top-level test that checks the number of triples after node load would be nice to have (separate PR)

node/src/gcp/value.rs
ppca: Yeah I don't want to mess up mpc code since it's working and I'm not sure maybe we retire mpc code base altogether someday so decided to write a new one.

ppca:
Do we have a mechanism to delete stale triples? Otherwise, our DB can grow indefinitely. 
-- yeah this is true. I could look into periodically cleaning the datastore table

integration-tests/tests/multichain/mod.rs
ppca: Yeah I wanted to do this but was not able to access the datastore correctly. The error is as shown a few lines down

ppca:
I assume that if some of the triples are not in sync between all the nodes, we will drop this particular protocol, choose the other 2 triples, and try again.
-- 
I think currently if one node goes offline while a protocol is in progress, then that protocol will be stuck forever. 
If the protocol for generating triples went thru successfully with all participants, but then I restarted without some triple, then other nodes could potentially try to use the triples I don't have to generate presignature, and that presignature protocol will be stuck forever.
However, as long as `presignature_manager.my_len() < 2 && presignature_manager.potential_len() < 10` satisfies, the node will continuously try to generate presignature by picking out distinct pairs (non-overlapping) of triples from triple_manager.mine, given there are >= 2 triples in triple_manager.mine. 
Triples will also be repeated generated if triple_manager.mine < 2 and total triples < 10.
So, in most cases, even if there are no synced triples among nodes (all triple_id in all nodes are different), the nodes will be able to generate new presignatures or generate new triples and use them for signature requests. However, there are corner cases like `triple_manager.my_len() == 1 yet triple_manager.len() >= 10` where the nodes will be stuck forever. I think @ChaoticTempest 's beacon triple generation might solve this as it continuous stock up triples.

volovyks:
> Do we have a mechanism to delete stale triples? Otherwise, our DB can grow indefinitely. -- yeah this is true. I could look into periodically cleaning the datastore table

This PR should be discussed together with https://github.com/near/mpc-recovery/pull/446. Check my last proposition, it may help with this issue,

node/src/protocol/triple.rs
```diff
@@ -247,25 +317,40 @@ impl TripleManager {
...
+        async fn new(number: u32) -> Self {
             let range = 0..number;
             // Self::wipe_mailboxes(range.clone());
-            let participants: Vec<Participant> = range.map(Participant::from).collect();
-            let managers = participants
-                .iter()
-                .map(|me| TripleManager::new(participants.clone(), *me, number as usize, 0))
+            let participants: Vec<Participant> = range.clone().map(Participant::from).collect();
+            let managers = range
+                .clone()
```
ChaoticTempest, line 340: hmm, why is this a `clone` instead of an `iter` or `into_iter`?

node/src/protocol/triple.rs
```diff
@@ -340,68 +425,70 @@ mod test {
...
-            tm.generate(0).unwrap();
-        }
-        tm.poke_until_quiet().unwrap();
-        tm.generate(1).unwrap();
-        tm.generate(2).unwrap();
-        tm.generate(4).unwrap();
-
-        tm.poke_until_quiet().unwrap();
+        let runtime = tokio::runtime::Runtime::new().unwrap();
+        runtime.block_on(async {
```
ChaoticTempest, line 432: you can also do a `#[tokio::test]` instead on top of the function just so the code here doesn't shift into a different scope

node/src/gcp/mod.rs
```diff
@@ -0,0 +1,398 @@
...
+
+        Ok(())
+    }
+}
+
+#[derive(Clone)]
+pub struct GcpService {
+    pub project_id: String,
+    pub datastore: Option<DatastoreService>,
+    pub secret_manager: Option<SecretManagerService>,
```
ChaoticTempest, line 333: why are these options? Would be unwieldy to use later when we go and directly reference them everytime?

node/src/protocol/consensus.rs
```diff
@@ -654,13 +675,24 @@ impl ConsensusProtocol for JoiningState {
...
-                Ok(NodeState::Started(StartedState(node_data)))
+                let persistent_node_data = ctx.secret_storage().load().await?;
+                let triple_storage = ctx.triple_storage();
+                let read_lock = triple_storage.read().await;
+                let triple_data_result = read_lock.load().await;
+                drop(read_lock);
+                let triple_data = match triple_data_result {
+                    Ok(vec_triple_data) => Some(vec_triple_data),
+                    _ => None,
+                };
```
ChaoticTempest, line 691: youu can shorten this to:
```suggestion
                let triple_data = match triple_data_result.ok();
```

node/src/protocol/state.rs
```diff
@@ -20,7 +21,10 @@ pub struct PersistentNodeData {
 }
 
 #[derive(Clone)]
-pub struct StartedState(pub Option<PersistentNodeData>);
+pub struct StartedState {
+    pub persistent_node_data: Option<PersistentNodeData>,
+    pub triple_data: Option<Vec<TripleData>>,
```
ChaoticTempest, line 26: why is this an option? I'd imagine this could just be an empty vec by default otherwise

ChaoticTempest (approved):

volovyks:
Nice work @ppca! Merging.

PR #446, opened 2024-02-06
feat: Added connection pool for offline participants

ChaoticTempest:
This creates a `connection::Pool` to be a single source of truth for all active participants. Note, that this isn't that clean and moreso to see if this API is good enough for us to go forward and use. It's the foundation for where we go next with networking.

This doesn't do anything fancy like TCP connections like how nearcore does it. Most of the logic is just calling into `Pool::ping` to get back the list of participants where we query for active participants every 5 seconds. This should more realistically be happening in a separate async task or thread of its own.


node/src/protocol/consensus.rs
```diff
@@ -160,6 +163,10 @@ impl ConsensusProtocol for StartedState {
                                 epoch,
                                 contract_state.old_epoch
                             );
+
```
volovyks, line 166: I know that the participants are not available in all states, but can we set them 1 time at the beginning of the advance function?

node/src/protocol/cryptography.rs
```diff
@@ -295,9 +297,10 @@ impl CryptographicProtocol for RunningState {
             tracing::warn!(?err, participants = ?self.participants, "running(pre): failed to send encrypted message");
         }
 
+        let participants = ctx.connection_pool().ping().await;
         let mut triple_manager = self.triple_manager.write().await;
         if triple_manager.my_len() < 2 && triple_manager.potential_len() < 10 {
-            triple_manager.generate()?;
+            triple_manager.generate(&participants)?;
```
volovyks, line 303: What will happen with triples that were generated when one of the participants was offline?
Should we use the connection pool only to kick members? If we generate triples for n-k participants while k  is still in the participant set, we can complicate the logic significantly.

volovyks (commented):
The implementation of the ConnectionPool is good, REST should be good for now.
I wonder if it's a good strategy in general. I have a feeling that the only source of truth for participants set should be in the contract.

node/src/protocol/consensus.rs
ChaoticTempest: I wonder if I can just generalize over the contract_state and set the participants like that, because there's some cases where the participants was never set, but doesn't seem like it's too much of an issue

node/src/connection.rs
```diff
@@ -0,0 +1,62 @@
...
+        if let Some((ref active, timestamp)) = *self.current_active.read().await {
+            if timestamp.elapsed() < DEFAULT_TIMEOUT {
+                return active.clone();
+            }
+        }
+
+        let connections = self.connections.read().await;
+        let mut participants = vec![];
+        for (participant, url) in connections.iter() {
+            let Ok(resp) = self.http.get(format!("{url}/state")).send().await else {
```
ppca, line 48: Do we need to set a timeout here? Would it hang forever?

node/src/protocol/cryptography.rs
ChaoticTempest: so, while the triple is getting generated here, we will have a list of active participants that were used to generate that triple. What hasn't been accounted for is when the node hops online again and doesn't have these triples nor do we check at all for the threshold. We will need some logic later where we account for these things such that the presignature will only use the participants from the triple generation participants. The big question here is whether that is secure or not

node/src/connection.rs
ChaoticTempest: The reqwest client should have a default 30 second timeout, so it should be fine

node/src/protocol/cryptography.rs
ppca: if triples were generated when one of the participants was offline:
1) if the participant was included in the set of participants to generate the triple but the participant went offline while triple protocol in progress, then the protocol will be stuck forever, and that triple never gets generated;
2) if the participant was already offline and came back up with a triple that's generated by other nodes. In this case, we need to make sure whenever we generate presignatures, we are only involving the set of participants that are involved in generating both of the 2 triples used for presignature. Similarly when we generate signature, we need to check the participants for the presignatures being used. But here we might have an issue: if the overlap participants between the 2 triples are already <= threshold, should we still proceed? I guess answer is no. We should perhaps set some limit to how big the participants set need to be to start a protocol for triples 🤔 like suppose we are k out of n signing, and we want to always have at least k+x in the participant set for triples, where 2*(k+x) > n + k, i.e. x > (n-k)/2 so there's always an overlap of > k participants in the presignature stage. We want to make sure we still have at least k participants in the final signature stage, otherwise we should choose another pair of triples. 🤔 

Looking at the complication of 2), it might be better

ChaoticTempest: I'd say we punt off adding any sort of kicking logic at all. If it becomes a huge problem, each of node can manually call vote_leave on a particular node. It's not the most important thing to consider right now for the beta. Kicking logic can be disastrous with the initial set of nodes we get for genesis where kicking the wrong node can potentially create a state where we have too few nodes to keep on doing resharing.

We should however be cleaning up the node state (triples, presigs) on resharing though.

ChaoticTempest:
No longer going to be doing cohorts, since we actually can use subsets of participants when it comes to presignatures and signatures where

- `presignature.participants` is a subset of `intersection(triple0.participants, triple1.participants)`
- `signature.participants` is a subset of `presignature.participants`.

What's not implemented is also do an intersection on active participants and participants of the triples and presignatures.

ChaoticTempest:
Scratch not having the intersection implementation. I added it anyways just so that this is more correct and guaranteed to work. I'll stop adding more things to this PR as I don't want to snowball it any further

volovyks (approved):
Looks good!

node/src/mesh/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+pub struct Mesh {
+    /// Pool of connections to participants. Used to check who is alive in the network.
+    pub connections: connection::Pool,
+
+    /// Participants that are active at the beginning of each protocol loop.
+    pub active_participants: Participants,
+
+    /// Potential participants that are active at the beginning of each protocol loop. This
+    /// includes participants belonging to the next epoch.
+    pub active_potential_participants: Participants,
```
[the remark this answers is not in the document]
ChaoticTempest: this is for when resharing happens. The active potential participants are from the new participants that are about to join. Currently not used due to the resharing protocol not taking in newly active participants so when we do `active.and(new_participants)`, we expect new participants to be online right now

volovyks:
@ChaoticTempest let's merge this and followup PRs, this one should make it to the testnet release with real partners.

volovyks (approved):

PR #451, opened 2024-02-09
fix: CI tests

ChaoticTempest:
This will make our tests use near-primitives 0.20.1 which should fix the CI compilation error. This also upgrades other dependencies as well up to the latest patch version


ChaoticTempest:
well this fix is quite painful. We may need to update all our near-* dependencies to the latest

PR #452, opened 2024-02-09
fix: easier fix for ci tests

ChaoticTempest:
Hopefully this one fixes the CI test more easily. Only needed to exclude near-sdk dependency from node/ repo. We should ideally add it back later but for now, let's just get tests running and working. Properly fixing it would require updating all dependencies to the latest: https://github.com/near/mpc-recovery/pull/451


volovyks:
We may get rid of `near_sdk` in the node completely by replacing near_sdk::PublicKey in the contract. (I was mentioning this before). The fix may work for now.

PR #453, opened 2024-02-10
Added key versions

DavidM-D:
Now we can change the root key if needed without needing to change the signing contracts API.

Not sure that's how you do an assert in contract, but I was going to get dragged in review anyway.


volovyks:
@DavidM-D we have too many open PRs. I've made the new field optional. If there is something you want to add - let's do that in the followup PR.

volovyks (approved):

PR #458, opened 2024-02-12
GCE Based Multichain

kmaus-near:
This is an initial test run at running Multichain on GCE Virtual Machines instead of cloudrun.


volovyks:
@kmaus-near can we fix the tests here? Is it ready for merging?

kmaus-near:
> @kmaus-near can we fix the tests here? Is it ready for merging?

It is ready to be merged, this environment is already live on dev and testnet. The terraform flow will not work with GCE based architecture and we should discuss development PR flow going forward.

.github/workflows/multichain-prod.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: 'google-github-actions/setup-gcloud@v2'
+  
+      - name: 'Set project'
+        run: 'gcloud config set project pagoda-discovery-platform-prod'
+
+      - name: 'Update Nodes'
+        run: |
+          gcloud compute instances update-container multichain-${{ github.event.inputs.network }}-0
+          gcloud compute instances update-container multichain-${{ github.event.inputs.network }}-1
+          gcloud compute instances update-container multichain-${{ github.event.inputs.network }}-2
```
volovyks, line 58: Let's fix this, there is no newline at the end of half the files.

infra/multichain-vm-dev/terraform-dev.tfvars
```diff
@@ -0,0 +1,40 @@
+env          = "dev"
+# project      = "pagoda-discovery-platform-dev"
```
volovyks, line 2: Do we need this commented code?

infra/modules/instance-from-tpl/metadata.yaml
```diff
@@ -0,0 +1,285 @@
+apiVersion: blueprints.cloud.google.com/v1alpha1
```
volovyks, line 1: What are these metadata files? They look generated. Do we need them in our codebase?

volovyks (commented):
We have a ton of config duplication. Can we avoid that?
Are there any configs or scripts we want to delete since we no longer use CLoudRun?

infra/modules/instance-from-tpl/metadata.yaml
kmaus-near: This is used as a template to send to google when creating an instance template, we do need this in the codebase

ppca (approved):

volovyks:
@kmaus-near please, address all the comments in the follow-up PR.

kmaus-near:
> We have a ton of config duplication. Can we avoid that? Are there any configs or scripts we want to delete since we no longer use CLoudRun?

we should probably keep them for the time being until we are fully off of cloudrun in all environments. We are in a transition state at the moment where MPC is still actively taking traffic on cloud run (I think) and multichain currently has no Mainnet version. Once we are ready to fully move away from cloudrun, then we can remove all of these cloudrun infrastructure components and workflow scripts.

infra/multichain-vm-dev/terraform-dev.tfvars
kmaus-near: removed

.github/workflows/multichain-prod.yml
kmaus-near: done

ppca (approved):

PR #461, opened 2024-02-13
Allow sabmitting the same poyload

volovyks:
- Added the test for sync the `sign` call
- Allowed submitting the same payload again. `near-fetch` retries and throws "Signature for this payload already requested". I'm still not sure how and why that works in NEAR CLI.

I do not think that check is important now, but it's causing many problems for testing and development. We will need to redesign it when we have more time.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -309,14 +309,10 @@ impl MpcContract {
             payload,
             path
         );
-        match self.pending_requests.get(&payload) {
-            None => {
-                self.pending_requests.insert(&payload, &None);
-                log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
-                Self::ext(env::current_account_id()).sign_helper(payload, 0)
-            }
-            Some(_) => env::panic_str("Signature for this payload already requested"),
-        }
+        // TODO: prevent submitting the same paload twice which can cause conflicts
+        self.pending_requests.insert(&payload, &None);
+        log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
+        Self::ext(env::current_account_id()).sign_helper(payload, 0)
     }
 
     #[private]
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -4,7 +4,7 @@ use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::collections::LookupMap;
 use near_sdk::log;
 use near_sdk::serde::{Deserialize, Serialize};
-use near_sdk::{env, near_bindgen, AccountId, PanicOnDefault, Promise, PromiseOrValue, PublicKey};
+use near_sdk::{env, near_bindgen, AccountId, PanicOnDefault, PromiseOrValue, PublicKey};
 use primitives::{CandidateInfo, Candidates, Participants, PkVotes, Votes};
 use std::collections::{BTreeMap, HashSet};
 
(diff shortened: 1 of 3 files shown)


contract/src/lib.rs
```diff
@@ -309,14 +309,10 @@ impl MpcContract {
...
-                self.pending_requests.insert(&payload, &None);
-                log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
-                Self::ext(env::current_account_id()).sign_helper(payload, 0)
-            }
-            Some(_) => env::panic_str("Signature for this payload already requested"),
-        }
+        // TODO: prevent submitting the same paload twice which can cause conflicts
+        self.pending_requests.insert(&payload, &None);
+        log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
+        Self::ext(env::current_account_id()).sign_helper(payload, 0)
```
ChaoticTempest, line 315: Let's split this off into several cases since we don't want to potentially erase the signature if it's already there:
```suggestion
        match self.pending_requests.get(&payload) {
            None => {
                self.pending_requests.insert(&payload, &None);
                log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
                Self::ext(env::current_account_id()).sign_helper(payload, 0)
            }
            Some(None) => Self::ext(env::current_account_id()).sign_helper(payload, 0),
            Some(Some(signature)) => PromiseOrValue::Value(signature),
        }
```

integration-tests/tests/multichain/actions/mod.rs
```diff
@@ -58,6 +59,45 @@ pub async fn request_sign(
     Ok((payload, account, tx_hash))
 }
 
+pub async fn request_sign_sync(
```
ChaoticTempest, line 62: we shouldn't need to test both versions though. They're pretty much the same, like there's no difference besides the calling mechanism. They're going to be hitting the same code path in near node side anyways

contract/src/lib.rs
ChaoticTempest: Additionally, let's just remove the removal of the pending signatures, such that this contract just caches the signatures for now

ChaoticTempest: Resolved with the commit I pushed

integration-tests/tests/multichain/actions/mod.rs
ChaoticTempest: We'll fix this one later

ChaoticTempest:
Maybe let's hold off merging this one since @DavidM-D is going to make a separate MPC service for people to make use of and test

integration-tests/tests/multichain/actions/mod.rs
volovyks: I understand that we are testing the same code from our perspective. It's more of a "let's check if MPC returns the signature in time" test. Can be done differently, but this one seems the ost convenient for me.

DavidM-D:
I might be reading this wrong, I think this might be a bit worse than what we have right now.

Since two people can be waiting on a response for the same payload, but they both will try to evict the response when they get it, only one of them (randomly selected) will ever succeed. This seems worse than the current solution of saying "you're too late" immediately to the second caller.

As discussed, I think we should either count the number of outstanding requests for a good fix.

Failing that, I'd remove the eviction of results (with a big TODO mentioning the memory leak it causes) which would allow both results to succeed.

volovyks:
Ok, I will invest more time into it once I have it.
My point was that the chances of getting the same payload are -> 0. (unless it's intentional)

DavidM-D:
OK this is causing @firatNEAR's team problems. If we can't get the dummy signatures working, let's do a hacky fix and work out how to do it right on Monday.

ChaoticTempest:
I already removed evicting the result, so we just basically have the signature cached for the lifetime of the contract now. Lets merge this then

PR #464, opened 2024-02-15
Added functions that generate dummy signatures

DavidM-D:
We want to unblock people when our service isn't working and take some pressure off keeping the damn thing up.

This PR hard codes in a private key with a value of 0 and a nonce with a value of 1 and does accurate key derivation on it. Provided you fetch the dummy root key, client code should behave consistently. The only small difference is I don't include episilon or a changing nonce meaning that signatures with an identical predecessor, derivation and payload will be identical.

This currently doesn't build on WASM and I suspect my moving around of dependencies will have bloated the size of the contract.


volovyks:
The build on my M1 fails as well. It's happening because of the `secp256k1` dependency (a dependency of `near-crypto` and `curv-kzen` crates). It's a c-binding library. Related issues [1](https://github.com/rustwasm/team/issues/291) and [2](https://github.com/rust-bitcoin/rust-secp256k1/issues/485).

On x86 Ubuntu it works, but the test is failing with 
```
let uniform_index = usize::from_le_bytes(usize_seed) % self.aliases.len();
     |                             -------------------- ^^^^^^^^^^ expected an array with a fixed size of 4 elements, found one with 8 elements
     |                             |
     |                             arguments to this function are incorrect
```

This is because `usize` is 4 bytes instead of 8 on 32-bit architecture. Looks like the `near-primitives` library is not designed for such use. The error is happening on `near-primitives-0.17.0/src/rand.rs:62:50`

One more related issue/NEP: https://github.com/near/NEPs/issues/443

ChaoticTempest:
hmm, I'm not sure what you guys are seeing, but this is compiling and the tests are running locally for me

ChaoticTempest:
> This currently doesn't build on WASM and I suspect my moving around of dependencies will have bloated the size of the contract.

The dependencies you added are a part of dev-dependencies, so that won't bloat the contract size at all

ChaoticTempest:
hmm, looks like the contract can't be deployed. I'll get back to this if I find some time, but feel free for anyone to try and solve it

volovyks:
@ChaoticTempest I was able to deploy it, but I'm getting `Instantiate` error.
The documentation says: `Error happened during instantiation.  This might indicate that `start` function trapped, or module isn't instantiable and/or unlinkable`.

volovyks:
@ChaoticTempest I've tried to refactor the contract, but faced other compilation issues. It's in `serhii/dummy-endpoints`.
We can try moving the derivation logic from https://github.com/near/multichain-gas-station-contract/blob/master/src/kdf.rs
It uses fewer dependencies and works (I have not tested it myself).

We can simplify it to a single `sign` function.

volovyks:
Looks like we do not need this logic anymore.

PR #470, opened 2024-02-28
fix/test: derive recovery id for same address

ChaoticTempest:
This adds in a test to derive the recovery id and check the same address can be recovered. The recovery id is derived by brute forcing whether or not the same public key can be recovered as the one we derive through MPC.


integration-tests/Cargo.toml
```diff
@@ -15,6 +15,7 @@ bollard = "0.13"
 cait-sith = { git = "https://github.com/LIT-Protocol/cait-sith.git", features = [
     "k256",
 ] }
+# cait-sith = { path = "../../cait-sith", features = ["k256"] }
```
DavidM-D, line 18:
```suggestion
```

node/Cargo.toml
```diff
@@ -15,6 +15,7 @@ aws-sdk-s3 = "0.24.0"
 aws-types = "0.54.0"
 axum = { version = "0.6.19" }
 axum-extra = "0.7"
+# cait-sith = { path = "../../cait-sith", features = ["k256"] }
```
DavidM-D, line 18:
```suggestion
```

DavidM-D (dismissed):
LGTM, with some small changes

node/src/kdf.rs
```diff
@@ -41,3 +46,58 @@ pub fn derive_delta(receipt_id: CryptoHash, entropy: [u8; 32]) -> Scalar {
 pub fn derive_key(public_key: PublicKey, epsilon: Scalar) -> PublicKey {
     (<Secp256k1 as CurveArithmetic>::ProjectivePoint::GENERATOR * epsilon + public_key).to_affine()
 }
+
+#[derive(Debug)]
+pub struct MultichainSignature {
```
ppca, line 51: should this be called MultichainEthSignature if only ETH is using the recovery id?

integration-tests/tests/multichain/actions/mod.rs
```diff
@@ -87,3 +101,293 @@ pub async fn single_signature_production(
...
+    assert_eq!(user_address_from_pk, recovered_from_signature_address_web3);
+
+    let recovered_from_signature_address_ethers = signature.recover(payload_hash_reversed).unwrap();
+    assert_eq!(
+        user_address_from_pk,
+        recovered_from_signature_address_ethers
+    );
+
+    let recovered_from_signature_address_local_function =
+        recover(signature, payload_hash_reversed).unwrap();
```
ppca, line 245: nit: We could delete the local recover() and the as_signature() used for local recover() since they do exactly same thing  as the recover() from the 2 crates, it just allowed me to print things out while debugging.

ppca (dismissed):

volovyks (approved):
Great update!

ppca (dismissed):

volovyks (approved):

volovyks (approved):

PR #480, opened 2024-03-08
feat: stop infinite loop of triple timeout by adding cache for failed triples

ppca:
**Purpose**
While trying out redeploying nodes at different times, I find that in the case when a large number of (>10) triple generation protocols are ongoing, the nodes could go into the infinite loop of: timing out the protocol for a triple id -> receiving message to generate the same triple from another node that has not timed out their protocol -> joining protocol to generate that same triple -> timing out.  This happens because each node would join the triple generation protocol at a different time and nodes cannot time out other nodes' protocols.
This PR adds a cache for failed triples, so that each node will not retry any failed (including timed out) triples. Then when the other nodes do not receive responses from this node about this triple protocol, they will time it out and thus close the loop for this triple generation instead of infinite looping. They could move on to generating other triples instead of resources being stuck in infinite loops of failed triples.

**What's changed in code**
1) add failed_triples: triple_id -> timestamp to TripleManager
2) when a triple generation protocol fail, add the triple_id and the timestamp we find out it failed to the failed_triples
3) when we received a message to generate a triple which has ID that is in failed_triples, we skip that message, and update the timestamp of that ID in failed_triples 
4) clear_failed_triples() will run at the end of run(), this clears all failed triples that either failed or messaged more than 2 hrs ago

### DIFF
--- a/node/src/protocol/message.rs
+++ b/node/src/protocol/message.rs
@@ -213,12 +213,12 @@ impl MessageHandler for RunningState {
         let participants = ctx.mesh().active_participants();
         let mut triple_manager = self.triple_manager.write().await;
 
-        // remove the triple_id that has been timed out from the triple_bins
+        // remove the triple_id that has already failed from the triple_bins
         queue
             .triple_bins
             .entry(self.epoch)
             .or_default()
-            .retain(|id, _| !triple_manager.triples_time_out.contains(&id));
+            .retain(|id, _| !triple_manager.failed_triples.contains_key(id));
 
         for (id, queue) in queue.triple_bins.entry(self.epoch).or_default() {
             if let Some(protocol) = triple_manager.get_or_generate(*id, participants)? {
@@ -312,6 +312,7 @@ impl MessageHandler for RunningState {
                 );
                 queue.extend(leftover_messages);
             }
+            triple_manager.clear_failed_triples();
         }
         Ok(())
     }
--- a/node/src/protocol/message.rs
+++ b/node/src/protocol/message.rs
@@ -15,6 +15,7 @@ use near_primitives::hash::CryptoHash;
 use serde::{Deserialize, Serialize};
 use std::collections::{HashMap, VecDeque};
 use std::sync::Arc;
+use std::time::Instant;
 use tokio::sync::RwLock;
 
 #[async_trait::async_trait]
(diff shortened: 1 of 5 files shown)


node/src/protocol/triple.rs
```diff
@@ -166,7 +175,7 @@ impl TripleManager {
             max_triples,
         } = self.triple_cfg;
         let not_enough_triples =
-            || self.my_len() < min_triples && self.potential_len() < max_triples;
+            || self.my_len() < min_triples && self.potential_len() - self.len() < max_triples;
```
ChaoticTempest, line 178: so doing `- self.len()` will get us `self.generators.len()`, which would mean that we would always be generating and have a max_triples amount of generators alive. Don't think we should do this `max_triples` should determine the total amount of triples in the whole network

ChaoticTempest (commented):
LGTM besides that one line

node/src/protocol/triple.rs
ppca: no this only generates when both mine < min_triples and total generators < max triples. This is necessary because there's an edge case where mine = 1 and total generated >= 10, this condition will successfully break that edge case.

ppca: So there's an edge case with mine = 1 and total = 10, and we will never generate any more triples for this case. And since presignatures need 2 mine, this 1 triple will never be used. I was doing above change to break this. But I guess a more reasonable way would actually be something like my_generator_len()

ChaoticTempest: I think we should have a different config then for generators, something like `max_concurrent_triples`, where
`min_triple < max_concurrent_triples < max_triples`. So `max_triples` still determines the total supply of triples in the network

node/src/protocol/message.rs
```diff
@@ -208,6 +209,20 @@ impl MessageHandler for RunningState {
...
+            .entry(self.epoch)
+            .or_default()
+            .retain(|id, _| {
+                let has_failed = triple_manager.failed_triples.contains_key(id);
+                if has_failed {
+                    triple_manager.failed_triples.insert(*id, Instant::now());
+                }
+                !has_failed
+            });
+
```
volovyks, line 225: Why we are not getting rid of the failed triple but adding failed triples?

ppca: We are getting rid of the failed triples. This logic refreshes the timestamp of the failed triple, and then only retains the triples that have not failed.

node/src/protocol/triple.rs
ppca: I changed condition to `(self.my_len() < min_triples && self.potential_len() == self.len())
                || self.potential_len() < max_triples`.
Basically generate triple when mine < min_triples and no triple protocol ongoing, or when existing + ongoing triples < max_triples.

ChaoticTempest (approved):

PR #490, opened 2024-03-13
feat: metrics for node up and signing latency

ppca:
whether node is up, sign request count, sign success, sign latency

### DIFF
--- a/node/src/indexer.rs
+++ b/node/src/indexer.rs
@@ -7,6 +7,7 @@ use near_lake_primitives::actions::ActionMetaDataExt;
 use near_lake_primitives::{receipts::ExecutionStatus, AccountId};
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
+use std::time::Instant;
 use tokio::sync::RwLock;
 
 /// Configures indexer.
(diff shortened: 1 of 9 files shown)


node/src/indexer.rs
```diff
@@ -120,7 +121,11 @@ async fn handle_block(
                             epsilon,
                             delta,
                             entropy,
+                            time_added: Instant::now(),
```
ChaoticTempest, line 124: nit: how about we keep it consistent with `timestamp` everywhere else since timestamp is usually associated to when something is created anyways

node/src/metrics.rs
```diff
@@ -0,0 +1,80 @@
+use once_cell::sync::Lazy;
+pub use prometheus::{
+    self, core::MetricVec, core::MetricVecBuilder, exponential_buckets, linear_buckets, Counter,
+    Encoder, Gauge, GaugeVec, Histogram, HistogramOpts, HistogramVec, IntCounter, IntCounterVec,
+    IntGauge, IntGaugeVec, Opts, Result, TextEncoder,
+};
+
+pub(crate) static NODE_RUNNING: Lazy<IntGaugeVec> = Lazy::new(|| {
```
ChaoticTempest, line 8: is it by convention to create `Lazy` values for doing metrics in prometheus? If not, we should have it be initialized from one place when the node starts up. It's hard to configure these things down the line since they're globals, but it's not something we have to do right now since there are more important things. If that's the case, then let's also create a ticket to revisit later

ChaoticTempest (approved):
LGTM

volovyks (approved):

PR #491, opened 2024-03-13
Contract reinit should not be possible

volovyks:


ChaoticTempest (approved):
LGTM, but I guess for testing, if we ever need to call a new init, we have to call into clean now

PR #492, opened 2024-03-14
Reject unregistered participants

volovyks:


volovyks:
@ppca we will need to update our and partner configs once this is merged

ppca:
If this allowlist is an ENV variable, then each time we update the allowlist, we'd need to ping the partners to update their node. It's probably OK to ping them, and then when all nodes have updated, the new participant would be able to join.
If we want this update to be automatic, we'd then need be ok with allowlist being public. Basically we specify the allowlist in code. Then every time we update the allowlist, we simply rebuild the docker image and update the the public image. And the watchtower inside each partner's VM will update their multichain node too.

volovyks:
A public list is not a bad thing. We can keep it in the contract that is controlled by Pagoda. But let's not forget about decentralization. Each partner should make their own decision on "who can join" because now the vote function is called automatically. cc @DavidM-D

ppca:
keeping it in the contract sounds like a good idea to me.

ChaoticTempest:
So the issue with this is that since we restart the node to reflect these changes, it might be bad to do so since the service should be running regardless of whether or not resharing is happening or not. So imagine all nodes restart at once to do this voting which isn't ideal for the liveliness of the service. So in that case, it's better to just have the CLI commands handy for people to use at that point. And let's just disallow people from joining arbitrarily for now then with this PR

volovyks:
We can get rid of that endpoint completely and ask node providers to use CLI, yes. But we will force them to keep the key locally on their PC.
The situation when all the nodes restart simultaneously is unlikely because people will update this list manually.

I do not have a strong opinion here, since all the options have downsides.

ChaoticTempest:
@volovyks so they can ssh into the node itself and then directly use the local key already stored there to do CLI call if that helps. Additionally, with the changes you have here with the allowlist, they would have to restart the node to do a manual CLI call

volovyks:
Closing in favour of https://github.com/near/mpc-recovery/pull/498

PR #493, opened 2024-03-14
Feat: added partner code for infrastructure, and updated dev code

kmaus-near:


infra/partner-vm/resources.tf
```diff
@@ -0,0 +1,41 @@
...
+      source  = "hashicorp/google"
+      version = "4.73.0"
+    }
+  }
+}
+
+# These data blocks grab the values from your GCP secret manager, please adjust secret names as desired
+data "google_secret_manager_secret_version" "account_sk_secret_id" {
+  count   = length(var.node_configs)
+  secret  = "multichain-account-sk-dev-${count.index}"
```
ppca, line 18: This actually forces the secret name to be multichain-account-sk-dev-x instead of giving partner freedom to choose their secret name? This is going to be problematic if the partner specifies some other name in node configs and store that into Secret manager.

infra/partner-vm/resources.tf
```diff
@@ -0,0 +1,41 @@
...
+  project = var.project_id
+}
+
+data "google_secret_manager_secret_version" "cipher_sk_secret_id" {
+  count   = length(var.node_configs)
+  secret  = "multichain-cipher-sk-dev-${count.index}"
```
ppca, line 24: same issue as above

infra/partner-vm/resources.tf
```diff
@@ -0,0 +1,41 @@
...
+  project = var.project_id
+}
+
+data "google_secret_manager_secret_version" "sk_share_secret_id" {
+  count   = length(var.node_configs)
+  secret  = "multichain-sk-share-dev-${count.index}"
```
ppca, line 30: same issue as above

infra/partner-vm/terraform-testnet-example.tfvars
```diff
@@ -0,0 +1,13 @@
+env          = "testnet"
+# These will be specific to your node
+node_configs = [
+  {
+    # Each node has a unique account ID
+    account              = "multichain-node-testnet-7.testnet"
+    cipher_pk            = "5a9d1d27fc3c952e7af7a2f1c84f552928eec232c3f2c3e787f4a15d82a82916"
```
ppca, line 7: let's change this to <your_cipher_pk>.

infra/partner-vm/variables.tf
```diff
@@ -0,0 +1,123 @@
...
+    value = string
+  }))
+  default = [
+    {
+      name  = "MPC_RECOVERY_NEAR_RPC"
+      value = "https://rpc.testnet.near.org"
+    },
+    {
+      name  = "MPC_RECOVERY_CONTRACT_ID"
+      value = "multichain0.testnet"
```
ppca, line 92: This likely will be some other contract for testnet. Let's update this when we decide that name.

ppca:
oh also, let's name the folder `partner-vm-testnet` instead of `partner-vm`. Some day someone will confuse mainnet and testnet and boom :(

infra/partner-vm/resources.tf
kmaus-near: They can update this name to be whatever they choose, I just wanted something there to show the format, but I can use the variable in the tfvars file maybe, I'll give it a try

kmaus-near: Updated to use tfvars

infra/partner-vm/terraform-testnet-example.tfvars
kmaus-near: done

ppca (approved):
LGTM. I may change a few places in followup PR so it's consistent with names I have partners pick in the [prep doc](https://docs.google.com/document/d/1xZplFvA16WpQUHAvpwoj9bMipvV5MShleMfmSzyCtzg/edit#heading=h.yrfxmjx3sd4f).

volovyks:
@kmaus-near is it ready to be merged?

volovyks:
@kmaus-near are we setting up any kind of firewall here? Will be a problem for new nodes that are joining the network?

kmaus-near:
> @kmaus-near are we setting up any kind of firewall here? Will be a problem for new nodes that are joining the network?

This instance is behind a load balancer, so the only firewall rule needed is the loadbalancer healthcheck which is configured at the bottom of main.tf. Since the load balancer will have a public static IP, nodes should not need any firewall rules to reach the instance.

kmaus-near:
> @kmaus-near is it ready to be merged?

Unless there's any objections from @ppca, I think we should be good with merging.

ppca (approved):

PR #499, opened 2024-03-19
fix: Added max_concurrent_generation for limiting triple timeouts

ChaoticTempest:
This adds queuing up triple generation and limiting to `max_concurrent_generation`. @ppca regardless of whether or not `max_concurrent_generation` is set, generating triples really takes a long time when the number of node increases. Local testing with 8 nodes is gonna be slow, so let me know if you were able to setup a network and try various `max_concurrent_generation` amounts to see if it works.

Otherwise, we'll need to parallelize triple generation itself into threads due to the resource contention and timing out. We could also raise the timeouts on the triple generation to something like 15mins instead of 5mins.


ppca:
let me know if you were able to setup a network and try various max_concurrent_generation amounts to see if it works. --- Yeah I can try it out easily on our dev env.

volovyks (approved):
LGTM, I will merge it to be able to do more precise tests.

PR #500, opened 2024-03-19
Keep the build caches between builds

DavidM-D:
When rustflags change all build artifacts are invalidated, because they are present in the build of all dependencies. This PR keeps the wasm builds and the native builds separate allowing them to maintain their caches. This may break some people's workflows if they're manually building and running the integration tests.


DavidM-D:
On the following benchmark

```rust
#[test(tokio::test)]
async fn overhead_bench() -> anyhow::Result<()> {
    with_multichain_nodes(MultichainConfig::default(), |ctx| {
        Box::pin(async move {
            panic!();
        })
    })
    .await
}
```
Before:
```
cargo test multichain::overhead_bench -- --exact --test-threads 1  619.91s user 25.71s system 602% cpu 1:47.19 total
```

After
```
cargo test multichain::overhead_bench -- --exact --test-threads 1  1.43s user 0.78s system 14% cpu 15.621 total
```

volovyks (approved):
Tested locally, works cool!
We will need caching on CI thought, it's still 30+ minutes.

PR #504, opened 2024-03-19
fix: added naive intersection check for bypassing threshold

ChaoticTempest:
This will naively pop and insert back triples and presigs where the intersection between their participants and the current active participants do not meet the threshold requirement. This is naive and can have bad worst time tradeoffs but better to have it than not for now


volovyks (approved):

volovyks:
I have executed the tests one more time, it's throwing `ExecutionError("Exceeded the prepaid gas.")`  in `test_signature_large_stockpile`.

ChaoticTempest:
@volovyks yeah it's probably because the GH actions instance is slow to be running 8 nodes at once

node/src/protocol/presignature.rs
```diff
@@ -246,6 +246,11 @@ impl PresignatureManager {
         self.presignatures.remove(&id)
     }
 
+    pub fn insert_mine(&mut self, presig: Presignature) {
+        self.mine.push_back(presig.id);
```
ppca, line 250: nice, so basically it's deprioritized

ppca (approved):
LGTM. we could optimize how to pick the triple/presignature that has the largest overlap later.

volovyks (dismissed):

ppca (approved):

PR #506, opened 2024-03-20
test: many node stockpile

ChaoticTempest:
This sets the stockpile test to have 8 nodes and have a minimum triple amount of 10 per node.


ChaoticTempest:
@ppca with this branch, the following configuration works really well locally, so if they're spun up into multiple nodes, it should behave just as well:
```
    const NODES: usize = 8;
    const THRESHOLD: usize = 4;
    const MIN_TRIPLES: usize = 10;
    const MAX_TRIPLES: usize = 2 * NODES * MIN_TRIPLES;

    let triple_cfg = TripleConfig {
        min_triples: MIN_TRIPLES,
        max_triples: MAX_TRIPLES,
        max_concurrent_introduction: 8,
        max_concurrent_generation: 32,
    };
```

volovyks (approved):

PR #509, opened 2024-03-21
Make the key version mandatory

DavidM-D:
We'll update key versions when we increase the security of our system, or periodically lower the amount of assets accessible to former validator cohorts. New key versions mean new keys.

Previously this field was optional, but if we leave it optional we have to make a decision to either interpret None as 0, the least secure key version, or continually update it to the latest one, changing peoples keys from under them.

Since there isn't a good answer I feel it's better to make them mandatory, so people at least have to think about them a bit.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -295,8 +295,7 @@ impl MpcContract {
 
     #[allow(unused_variables)]
     /// `key_version` must be less than or equal to the value at `latest_key_version`
-    pub fn sign(&mut self, payload: [u8; 32], path: String, key_version: Option<u32>) -> Promise {
-        let key_version = key_version.unwrap_or(0);
+    pub fn sign(&mut self, payload: [u8; 32], path: String, key_version: u32) -> Promise {
         let latest_key_version: u32 = self.latest_key_version();
         assert!(
             key_version <= latest_key_version,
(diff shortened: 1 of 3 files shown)


DavidM-D:
Up to you guys if you want to merge this or not, it's your codebase.

Proximity is fine with it if it lands today, otherwise we can punt it until post launch.

contract/src/lib.rs
```diff
@@ -295,8 +295,7 @@ impl MpcContract {
 
     #[allow(unused_variables)]
     /// `key_version` must be less than or equal to the value at `latest_key_version`
-    pub fn sign(&mut self, payload: [u8; 32], path: String, key_version: Option<u32>) -> Promise {
-        let key_version = key_version.unwrap_or(0);
+    pub fn sign(&mut self, payload: [u8; 32], path: String, key_version: u32) -> Promise {
```
volovyks, line 298: Let's add key_version to log

volovyks (approved):

volovyks:
@ppca Once we release a new version, let's post in a chat that the key version is now mandatory

ppca:
@volovyks  the contract is not initialized yet, do we want to directly include this in the testnet launch?

DavidM-D:
I think so, I'd imagine that the nodes are running the version expecting this field?

volovyks:
@ppca once we have N (probably 8 nodes) we can test the automatic deployment of the new Docker image with this and other changes. We can init the contract after that.

PR #510, opened 2024-03-21
Multichain load tests

volovyks:
Load tests can be executed with
`cd load-tests`
`RUST_LOG=info cargo run --release -- --host https://rpc.testnet.near.org --report-file=load_test_results.html --test-plan "$(cat ./src/multichain/test_plans/short.txt)" --scenarios multichainSign`

The new code is in `load-tests/src/multichain`.

You can play with the test-plan file. For example:
`0,0s;1,120s;0,0s`
will create 1 user that will make calls one after another for 120 seconds.
You can set the number of users. All of them will use the same near acc for now, but in our case, it does not matter.


load-tests/src/multichain/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+    let jsonrpc_client = connector.connect(&testnet_rpc_url);
+    let rpc_client = near_fetch::Client::from_client(jsonrpc_client.clone());
+
+    let (nonce, block_hash, _) = rpc_client
+        .fetch_nonce(&signer.account_id, &signer.public_key)
+        .await
+        .unwrap();
+
+    let payload: [u8; 32] = rand::thread_rng().gen();
+    let payload_hashed = web3::signing::keccak256(&payload);
```
DavidM-D, line 40: The hash of a random [u8; 32] is just another random [u8; 32]
```suggestion
    let payload_hashed: [u8; 32] = rand::thread_rng().gen();
```

DavidM-D (approved):
LGTM

volovyks:
@DavidM-D I've accepted your commit and now I need 1 more approval :/
Do we really need this?

DavidM-D (approved):
If you work out who made the rule, I'll hassle them about it?

PR #512, opened 2024-03-21
No delay experiment

volovyks:
This change allowed me to run tests 2 times faster. All the infra setup time is the same, so that is probably 4x+ in performance for the protocol itself. The downside is we will have more requests to RPC.

2 and 100ms are chosen empirically. Gives me the same results as 0 delay.

(tested with 3 nodes, but should help with our scaling problem)


DavidM-D:
Why did we have that 1000ms delay in the first place?

volovyks:
@DavidM-D Without any delay Node will make too many RPC calls. Reading logs on the flight is also a bit easier.

node/src/protocol/mod.rs
```diff
@@ -255,6 +256,9 @@ impl MpcSignProtocol {
                 let guard = self.state.read().await;
                 guard.clone()
             };
+
+            let before_step = SystemTime::now();
```
ChaoticTempest, line 260: we should use `Instant::now()` and then call into `elapsed() < Duration::from_millis(10)` instead since that's more readable. 10ms is also for good measure since the amount of time it takes can vary based on different machines. Maybe even 50ms is better option.

node/src/protocol/mod.rs
```diff
@@ -277,11 +281,18 @@ impl MpcSignProtocol {
...
             *guard = state;
             drop(guard);
 
-            tokio::time::sleep(Duration::from_millis(1000)).await;
+            let step_duration = after_step.duration_since(before_step).unwrap().as_millis();
+
+            // If there was no work to do - sleep to avoid CPU and RPC burn
+            if step_duration < 2 {
+                tokio::time::sleep(Duration::from_millis(100)).await;
+            }
```
ChaoticTempest, line 295: we need similar logic for the error cases as well when they're sleeping for 1s

ChaoticTempest (commented):
Looking at this again, I'm not sure if we should add in this change now, since the protocol state advancement and the triple generation protocol should have separate sleep times. State advancement rarely ever happens so 1s is pretty ideal, meanwhile we want each triple protocol poke to advance much faster than just 1s. How about we do a simple check whether or not the protocol state is in the running state? If it goes from Running to Running state within `advance`, `progress`, `message.handle` then we can set the sleep to be nonexistent, but if we're in the Generating, WaitingForConsensus, Resharing state, we'll have the 1s sleep instead.

node/src/protocol/mod.rs
volovyks: Good point, changed

node/src/protocol/mod.rs
volovyks: I've set it to 100ms

volovyks:
I've changed the design.
On a top level, we want to exchange and process messages ASAP, in all states. At the same time, we do not want to read the contract state every X milliseconds, it should not change that often and we should not DDoS RPC.
I think the best option here is to have 0 delay for protocol and message processing and 1s delay for consensus. This strategy gave me the same results as no delays at all.

node/src/protocol/mod.rs
```diff
@@ -212,27 +213,14 @@ impl MpcSignProtocol {
...
-                    continue;
-                }
-            };
-            tracing::debug!(?contract_state);
             loop {
                 let msg_result = self.receiver.try_recv();
                 match msg_result {
                     Ok(msg) => {
-                        tracing::debug!("received a new message");
+                        tracing::debug!(?msg, "received a new message");
```
ChaoticTempest, line 223: this log here likely spams our logs. Do you think we need? It's not like we need to see what type of message we're getting though

ppca (approved):
LGTM

node/src/protocol/mod.rs
```diff
@@ -246,42 +234,61 @@ impl MpcSignProtocol {
...
                     continue;
                 }
             };
+
             if let Err(err) = state.handle(&self, &mut queue).await {
                 tracing::info!("protocol unable to handle messages: {err:?}");
-                tokio::time::sleep(Duration::from_millis(1000)).await;
+                tokio::time::sleep(Duration::from_millis(100)).await;
                 continue;
             }
```
ChaoticTempest, line 255: we should move this to after `state.advance` just so the messages for state advancement are handled before we assign `state` to `self.state`

ChaoticTempest (commented):
mostly looks good, just need to move `handle`

ChaoticTempest (approved):

PR #517, opened 2024-03-23
fix: signature timeout and add triple latency metric

ppca:


node/src/cli.rs
```diff
@@ -53,18 +53,18 @@ pub enum Cli {
         #[clap(flatten)]
         storage_options: storage::Options,
         /// At minimum, how many triples to stockpile on this node.
-        #[arg(long, env("MPC_RECOVERY_MAX_TRIPLES"), default_value("2"))]
+        #[arg(long, env("MPC_RECOVERY_MIN_TRIPLES"), default_value("20"))]
         min_triples: usize,
         /// At maximum, how many triples to stockpile on this node.
-        #[arg(long, env("MPC_RECOVERY_MAX_TRIPLES"), default_value("10"))]
+        #[arg(long, env("MPC_RECOVERY_MAX_TRIPLES"), default_value("120"))]
```
ChaoticTempest, line 59: should be at least a factor of 10 for the max. So something like 400 would be better if this is our prod value going forward

node/src/protocol/signature.rs
```diff
@@ -141,7 +143,8 @@ pub struct FailedGenerator {
     pub msg_hash: [u8; 32],
     pub epsilon: Scalar,
     pub delta: Scalar,
-    pub timestamp: Instant,
+    pub sign_request_timestamp: Instant,
```
ChaoticTempest, line 146: hmm doesn't look like sign_request_timestamp ever gets used. Maybe we just delete the value entirely

ChaoticTempest (approved):

node/src/protocol/signature.rs
ppca: sign_request_timestamp was used to pass to the generator created afterwards. The generator_timestamp field was not necessary so deleted.

ChaoticTempest (approved):

PR #518, opened 2024-03-23
Switched over to a newer version of cait sith

DavidM-D:
Lot's of security fixes and I believe triple generation messages should be 50% smaller.


volovyks (approved):
LGTM

PR #519, opened 2024-03-25
add more metric

ppca:
I added a field my_account_id to PresignatureManager so that it's easier for me to add metrics

### DIFF
--- a/node/src/metrics.rs
+++ b/node/src/metrics.rs
@@ -61,6 +61,106 @@ pub(crate) static TRIPLE_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
     .unwrap()
 });
 
+pub(crate) static PRESIGNATURE_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_latency_sec",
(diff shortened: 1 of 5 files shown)


node/src/metrics.rs
```diff
@@ -61,6 +61,106 @@ pub(crate) static TRIPLE_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
     .unwrap()
 });
 
+pub(crate) static PRESIGNATURE_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_latency_sec",
+        "Latency of multichain presignature generation, start from starting generation, end when presignature generation complete.",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
```
volovyks, line 69: Why do we need this?

volovyks (approved):
LGTM

node/src/metrics.rs
ppca: because the latencies are grouped into buckets, in this case it's buckets exponentials, [1,1.5), [1.5, 2.25) and so on, the last bucket will cover +inf. We could also do linear buckets if we want, but that will not cover +inf.

PR #520, opened 2024-03-25
Contract call limits

volovyks:


contract/src/lib.rs
```diff
@@ -1,7 +1,7 @@
 pub mod primitives;
 
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
-use near_sdk::collections::LookupMap;
+use near_sdk::collections::TreeMap;
```
ppca, line 4: what's the difference between the TreeMap and LookupMap?

ppca:
I tried redeploying this contract on dev env, and called state() fn, and got following error:
```
kind: {
    ExecutionError: 'Smart contract panicked: panicked at /Users/xiangyiz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/near-sdk-5.0.0-alpha.2/src/environment/env.rs:921:46:\n' +
      'Cannot deserialize the contract state.: Custom { kind: InvalidData, error: "Unexpected length of input" }'
  },
```
it went away after I redeployed using deploy branch code.

contract/src/lib.rs
volovyks: The lookup map does not know what is inside of it. It just stores elements directly using the formula:
`map_tag` + `<element key>`. TreeMap builds a tree of elements. In our case, LookupMap is more efficient, but probably it's not that important, since the max size of the map is 8.

volovyks:
@ppca that is probably because of the old state. Ideally, we should migrate, but it should not be a problem for now.

contract/src/lib.rs
ChaoticTempest: but what specifically do we get out of using `TreeMap` over `LookupMap`? Doesn't seem clear to me what the benefit here is

contract/src/lib.rs
```diff
@@ -337,6 +340,10 @@ impl MpcContract {
                     PromiseOrValue::Value(signature)
                 }
                 None => {
+                    if depth > 30 {
```
ChaoticTempest, line 343: why specifically 30? The max theoretical depth is ~60 and the max seen has been around late 50s. So why not 50?

volovyks: I've checked several transactions, and most of them ranged from 35 to 45. So I picked a reliable number. We do not want to have "out of gas" errors anymore. In good conditions we are returning a signature on step 8.

contract/src/lib.rs
volovyks: It is not possible to know the size of the `LookupMap`.

ppca (approved):

PR #530, opened 2024-03-26
add metrics around success rates

ppca:


node/src/protocol/triple.rs
```diff
@@ -183,10 +194,17 @@ impl TripleManager {
             self.me,
             self.threshold,
         )?);
+        let mine = true;
         self.generators
-            .insert(id, TripleGenerator::new(id, participants, protocol));
+            .insert(id, TripleGenerator::new(id, participants, protocol, mine));
```
ChaoticTempest, line 199: TripleManager::generate doesn't necessarily generate a `mine` triple. This is determined randomly so we can't do it like this

ChaoticTempest (approved):

ChaoticTempest (approved):

PR #531, opened 2024-03-26
Request counter

volovyks:


contract/src/lib.rs
```diff
@@ -49,7 +49,8 @@ pub enum ProtocolContractState {
 #[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
 pub struct MpcContract {
     protocol_state: ProtocolContractState,
-    pending_requests: TreeMap<[u8; 32], Option<(String, String)>>,
+    pending_requests: LookupMap<[u8; 32], Option<(String, String)>>,
+    request_counter: u32,
```
ppca, line 53: this will break contract I'm guessing? since when it was initialzed it did not have this field.

volovyks: It depends on how borsh deserialization works, I have not tested it. I belive it should be initialized on the first `sign` call.

ChaoticTempest: here's an example of how we can migrate state: https://github.com/near/near-sdk-rs/blob/1408127feefb37e7f84eba25984af18af46cd91d/examples/test-contract/src/lib.rs#L20

ChaoticTempest: so we need a new function similar to `clean` but ignores the state

ppca: Let me try redeploy the dev contract.

ppca: yep. I got this error ```
kind: {
    ExecutionError: 'Smart contract panicked: panicked at /Users/xiangyiz/.cargo/registry/src/index.crates.io-6f17d22bba15001f/near-sdk-5.0.0-alpha.2/src/environment/env.rs:921:46:\n' +
      'Cannot deserialize the contract state.: Custom { kind: InvalidData, error: "Unexpected length of input" }'
  },
```

ppca: all 3 contracts migrated

volovyks: k, migration function deleted

ppca (approved):

ChaoticTempest (approved):
We should move to versioned contracts so this process is a lot easier: https://github.com/near/mpc-recovery/issues/545

PR #536, opened 2024-03-27
Added init_running contract function

volovyks:


contract/src/lib.rs
```diff
@@ -72,6 +73,36 @@ impl MpcContract {
...
+            protocol_state: ProtocolContractState::Running(RunningContractState {
+                epoch,
+                participants: Participants { participants },
+                threshold,
+                public_key,
+                candidates: Candidates::new(),
+                join_votes: Votes::new(),
+                leave_votes: Votes::new(),
+            }),
+            pending_requests: TreeMap::new(b"m"),
```
ppca, line 102: what happens to the pending requests in old contract?

volovyks: We should not move old requests to the new contract since they will be outdated while you transfer them. Nothing bad will happen if nodes try to push signatures for a non-existent payload.
Votes are not transferred either.

ChaoticTempest: how would they be outdated? I think it's still fair to maintain these requests though otherwise clients will not be able to tell where their signature went (especially when the NEP lands). We should at the very least report back an error somehow

volovyks: They will get an error from the old contract. These signatures are short-lived and contract migration happens rarely. Also, nodes will need to change the contract manually and that will take more than 8 seconds. We will have some downtime.

volovyks: We may have a more complex design, like a voting mechanism for the "next" contract. But I do not think that is necessary now. This function is mostly for emergencies.

volovyks: About moving requests from contract to contract. That is not possible with the current design. We are targeting Sync calls only. Each request must be deleted from the queue in the same sign it was added with. Manually added requests will stay there forever.

ChaoticTempest: yeah this is fine for now as a emergency lever, but ideally we shouldn't be changing contracts. For the error, I'm talking about not just contract-level errors, but applications that are currently relying on indexer to check whether or not a signature has come in. Those applications will forever wait, which isn't good

volovyks: With the current design, nobody will wait for a signature in an async way. Even if we change the design in the future, such applications should have their timeout.

I agree that we should avoid moving from contract to contract since that is a breaking change for all the clients.
We should be able to handle all the breaking changes in the same contract without moving to a new one. 

When adding this function, I was mostly thinking about situations, when we lock the contract to make sure it can not be updated. And then if needed, we can move to another contract using this function.

Should we introduce another contract that will redirect sign calls to the current version of the real MPC contract?
We will control such a contract by setting the current MPC contract ID each time we change it. Loosing this key will not introduce any security threats.
At the same time, all versions of the MPC contract will be locked to prevent all sorts of attacks with stolen contract keys.

ppca (approved):

ChaoticTempest (approved):

PR #537, opened 2024-03-27
modify default timeout for triple = 20 min

ppca:


volovyks (approved):

ChaoticTempest (approved):
I or we should probably make these configurable parameters just so we don't have to recompile this every single time for deployments

ppca:
> I or we should probably make these configurable parameters just so we don't have to recompile this every single time for deployments

We can put it into ENv vars in the terraform code. Then we can redeploy by update-container command.

PR #549, opened 2024-03-30
feat: /msg endpoint now takes vectorized encrypted messages

ChaoticTempest:
This optimizes the number of messages outflowing from a node, such that the network doesn't get used as much. This bundles up multiple messages into at most 256kb payloads. Reduces the latency quite a bit and increases the success rate of triples generated in time


volovyks (approved):
What is the average number of messages in one 256kb partition?
We can fine-tune it once we switch to TCP, but why 256kb?

PR #550, opened 2024-03-30
chore: cleanup triple storage and unnecessary clones

ChaoticTempest:
Every time a triple was deleted, we needed to make unnecessary clones so this PR gets rid of that. Also further reduces unnecessary clones in our node/ code.


node/src/storage/triple_storage.rs
```diff
@@ -14,6 +14,25 @@ use std::collections::{HashMap, HashSet};
...
+    pub triple_id: TripleId,
+    pub mine: bool,
+}
+
+impl Keyable for TripleKey<'_> {
+    fn key(&self) -> Key {
+        Key {
+            path: Some(vec![PathElement {
+                kind: Some(TripleData::kind()),
+                name: Some(format!("{}/{}", self.account_id, self.triple_id)),
```
volovyks, line 28: Should we add mine here?

volovyks (approved):

node/src/storage/triple_storage.rs
ChaoticTempest: nope, mine does not need to be a part of the entity key. All triples are stored regardless of whether they belong to the node or not, where we have another field for `mine: bool`. This here is the key in reference of that entity, so it's not needed to store this value in the key

volovyks (approved):

PR #551, opened 2024-04-01
clear_payloads func

volovyks:
One of our envs is stuck because it has >8 payloads. PPPro because it had 8 payloads in the past.  I've added a clean function that should be called when this new version of the contract is deployed. @ppca please, do that for all our envs. And for `v5.multichain-mpc-dev.testnet` specifically.

@ChaoticTempest I've tried to add a read/write lock to requests, but borsh was not happy about that. Not sure if that is necessary at all.


contract/src/lib.rs
```diff
@@ -392,6 +392,12 @@ impl MpcContract {
         }
     }
 
+    #[private]
+    pub fn clean_payloads(&mut self) {
+        log!("clean_payloads");
+        self.pending_requests = LookupMap::new(b"m");
```
ChaoticTempest, line 398: were you able to test that this works? This actually doesn't do anything, since the state tree requires explicit modification, as in we need to explicitly call remove on each of the entries

ChaoticTempest:
> I've tried to add a read/write lock to requests, but borsh was not happy about that. Not sure if that is necessary at all.

How were you using locks? Locks in contract code should not be required at all. Also, you can't serialize a lock unless it's a filelock but that shouldn't be used here in contract code either

contract/src/lib.rs
volovyks: totally forgot about that, fixed

volovyks:
@ChaoticTempest yep, just looking for a reason why the counter can be off. Looks like resetting should help.

contract/src/lib.rs
```diff
@@ -423,4 +409,28 @@ impl MpcContract {
...
+        }
+        Self {
+            protocol_state: ProtocolContractState::NotInitialized,
+            pending_requests: LookupMap::new(b"m"),
+            request_counter: 0,
+        }
+    }
+
+    #[private]
+    pub fn clean_payloads(&mut self, payloads: Vec<[u8; 32]>, counter: u32) {
```
ppca, line 429: why specify the payloads instead of clearing all the requests in the pending_requests?

ppca: I've used ```
#[private]
    pub fn clean_payloads(&mut self) {
        log!("clean_payloads");
        self.pending_requests = LookupMap::new(b"m");
        self.request_counter = 0;
    }
```
to clean the payloads on dev for now.

Comment (volovyks) [reply contract/src/lib.rs:429]: Ok, that is fine. But the old payloads are still there.
LookupMap can not clean its key-value pairs. It is optimized for storage and access speed.

**Comment by volovyks** [comment]
@ChaoticTempest @ppca ready for review

File: contract/src/lib.rs (hunk 2 above)
Comment (ppca) [reply contract/src/lib.rs:475]: wait, why would the old payloads still be there? I thought `self.pending_requests = LookupMap::new(b"m");
        self.request_counter = 0;` will reset it?

Comment (volovyks) [reply contract/src/lib.rs:475]: This is how insert and remove works in `LookupMap`
```rust
    /// Inserts a serialized key-value pair into the map.
    /// If the map did not have this key present, `None` is returned. Otherwise returns
    /// a serialized value. Note, the keys that have the same hash value are undistinguished by
    /// the implementation.
    pub fn insert_raw(&mut self, key_raw: &[u8], value_raw: &[u8]) -> Option<Vec<u8>> {
        let storage_key = self.raw_key_to_storage_key(key_raw);
        if env::storage_write(&storage_key, value_raw) {
            Some(env::storage_get_evicted().unwrap())
        } else {
            None
        }
    }

    /// Removes a serialized key from the map, returning the serialized value at the key if the key
    /// was previously in the map.
    pub fn remove_raw(&mut self, key_raw: &[u8]) -> Option<Vec<u8>> {
        let storage_key = self.raw_key_to_storage_key(key_raw);
        if env::storage_remove(&storage_key) {
            Some(env::storage_get_evicted().unwrap())
        } else {
            None
        }
    }
```

LookupMap is not aware of its key-value pairs. They are stored directly on the contract memory which is also a map. The key for storing values is "LopkupMap prefix + provided key". It is not possible to clear the Lookup map without passing the keys.

ChaoticTempest (approved):
lgtm

PR #552, opened 2024-04-01
Subaccounts for load tests

volovyks:
@ChaoticTempest I was not able to use Workspaces `create_tla` on testnet. But the subaccount approach works just fine.


ChaoticTempest:
Did you follow all the rules for creating a Top level account? You could just use `dev_create_account()` to facilitate this without needing to know the rules, but both functions uses the testnet faucet which can still be drained

load-tests/src/fastauth/mod.rs
```diff
@@ -29,14 +30,35 @@ use near_primitives::{
...
+
+    let worker = near_workspaces::testnet().await.unwrap();
+
+    let root_account = Account::from_secret_key(
+        near_workspaces::types::AccountId::try_from("dev-1660670387515-45063246810397".to_string()).unwrap(),
+        near_workspaces::types::SecretKey::from_str(
+            "ed25519:4hc3qA3nTE8M63DB8jEZx9ZbHVUPdkMjUAoa11m4xtET7F6w4bk51TwQ3RzEcFhBtXvF6NYzFdiJduaGdJUvynAi"
+        ).unwrap(),
+        &worker
+    );
```
ChaoticTempest, line 46: the issue with this is that somebody can easily disrupt the load test by getting this key. We wouldn't notice this either until we run it locally manually. We should be creating a dev account each time and delete the account at the end of the test

volovyks: Yep, I know. But `dev_create_account` does not work for me. And there is [no Error handling](https://github.com/near/near-workspaces-rs/issues/353).

volovyks:
@ChaoticTempest what kind of rules? Here is what I'm doing:
```rust
let worker = near_workspaces::testnet().await.unwrap();
let w_account = worker.dev_create_account().await.expect("Failed to create account");
tracing::info!("Created worker account: {}", w_account.id());
```
It's not throwing any errors, but the account is never created on testnet.

volovyks:
@ChaoticTempest looks like subaccounts is a way to go for now: https://github.com/near/near-workspaces-rs/issues/354

volovyks:
@ChaoticTempest added delete account step

ChaoticTempest (approved):

ppca (approved):

PR #553, opened 2024-04-02
Node and contract version metrics

volovyks:
The contract function is tested manually. For the node version, we will need to deploy and update Grafana config.
@ppca let's make sure we are [updating](https://semver.org/) the version before each deploy of contract and node.

### DIFF
--- a/node/src/protocol/mod.rs
+++ b/node/src/protocol/mod.rs
@@ -219,12 +219,9 @@ impl MpcSignProtocol {
         crate::metrics::NODE_RUNNING
             .with_label_values(&[&my_account_id])
             .set(1);
-        let mpc_contract_version = get_contract_version(&self.ctx.mpc_contract_id);
-        if let Ok(version) = mpc_contract_version {
-            crate::metrics::MPC_CONTRACT_VERSION
-                .with_label_values(&[&my_account_id])
-                .set(version);
-        }
+        crate::metrics::NODE_VERSION
+            .with_label_values(&[&my_account_id])
+            .set(node_version());
         let mut queue = MpcMessageQueue::default();
         let mut last_state_update = Instant::now();
         let mut last_pinged = Instant::now();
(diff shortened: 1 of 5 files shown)


node/src/protocol/mod.rs
```diff
@@ -359,9 +356,9 @@ async fn get_my_participant(protocol: &MpcSignProtocol) -> Participant {
     participant_info.id.into()
 }
 
-fn get_contract_version(mpc_contract_account: &AccountId) -> Result<i64, std::num::ParseIntError> {
-    let mpc_contract_id = mpc_contract_account.as_str();
-    let parts: Vec<&str> = mpc_contract_id.split('.').collect();
-    let version_str = parts[0].trim_start_matches('v');
-    version_str.parse::<i64>()
+fn node_version() -> i64 {
+    env!("CARGO_PKG_VERSION")
```
ppca, line 360: this is the version in 
```
fn node_version() -> i64 {

    env!("CARGO_PKG_VERSION")

        .split('.')

        .map(|s| s.parse::<i64>().unwrap())

        .fold(0, |acc, x| acc * 1000 + x)

}
```?

ppca:
`we are [updating](https://semver.org/) the version before each deploy of contract and node.` 
--
Does that mean each time before we deploy, we need to merge a PR with version change, and cut the deploy at that commit? I think testnet and mainnet is good to adopt this, but not sure about dev.

node/src/protocol/mod.rs
volovyks: @ppca I'm not sure what you are asking. Can you please clarify?

ppca: Ah I pasted the wrong thing. I mean this is the node/ version?

volovyks: This function will give us an integer representation of the version specified in `Cargo.toml`.

ppca (approved):

PR #554, opened 2024-04-02
reverse payload on the node

volovyks:
@DavidM-D does not seem like it works


volovyks:
@DavidM-D we will not reverse it

PR #557, opened 2024-04-03
Feat: dev deployment pipeline

### LINKED ISSUE #525: Automatic deployment of Dev env on merge to develop

kmaus-near:
Closes #525

### DIFF
--- a/.github/workflows/multichain-dev.yml
+++ b/.github/workflows/multichain-dev.yml
@@ -11,6 +11,7 @@ env:
 
 jobs:
   build-mpc-recovery:
+    if: github.event.pull_request.merged == true
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


.github/workflows/multichain-dev.yml
```diff
@@ -0,0 +1,59 @@
+name: Deploy Multichain Dev.
+on:
+  pull_request:
+    types: [closed]
```
ppca, line 4: should we differentiate between closed and merged?

ppca (approved):
one comment only about PR type

.github/workflows/multichain-dev.yml
kmaus-near: That's what line 14 and 34 do 😄

volovyks (commented):
LGTM, but let's add contract deployment, it would be very inconvenient to do that manually all the time.

ppca:
> LGTM, but let's add contract deployment, it would be very inconvenient to do that manually all the time.

yep. contract is coming up in another PR

PR #558, opened 2024-04-03
contract: make multichain contract versioned

ppca:
This will be our contract
```
pub enum VersionedMpcContract {
    V0(MpcContract),
}
```
I put majority of the fn logics in VersionedMpcContract implementation, and added functions that read data from contract, e.g.
```
fn mutable_state(&mut self) -> &mut ProtocolContractState {
        match self {
            Self::V0(ref mut mpc_contract) => &mut mpc_contract.protocol_state,
        }
    }

fn get_signature_per_payload(&self, payload: [u8; 32]) -> Option<Option<(String, String)>> {
        match self {
            Self::V0(mpc_contract) => mpc_contract.pending_requests.get(&payload),
        }
    }

    fn remove_request(&mut self, payload: &[u8; 32]) {
        match self {
            Self::V0(mpc_contract) => {
                mpc_contract.remove_request(payload);
            }
        }
    }

    fn add_request(&mut self, payload: &[u8; 32], signature: &Option<(String, String)>) {
        match self {
            Self::V0(mpc_contract) => {
                mpc_contract.add_request(payload, signature);
            }
        }
    }

    fn check_and_insert_request(&mut self, payload: [u8; 32], big_r: String, s: String) {
        match self {
            Self::V0(mpc_contract) => {
                mpc_contract.check_and_insert_request(payload, big_r, s);
            }
        }
    }
```
So each version could have different format of storing data and we could add code to above fn to read data from each version.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -49,6 +49,10 @@ pub enum ProtocolContractState {
 }
 
 #[near_bindgen]
+pub enum VersionedMpcContract {
+    V0(MpcContract),
+}
+
 #[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
 pub struct MpcContract {
     protocol_state: ProtocolContractState,
@@ -56,60 +60,42 @@ pub struct MpcContract {
     request_counter: u32,
 }
 
-#[near_bindgen]
 impl MpcContract {
-    #[init]
-    pub fn init(threshold: usize, candidates: BTreeMap<AccountId, CandidateInfo>) -> Self {
-        log!(
-            "init: signer={}, treshhold={}, candidates={}",
-            env::signer_account_id(),
-            threshold,
-            serde_json::to_string(&candidates).unwrap()
-        );
-        MpcContract {
-            protocol_state: ProtocolContractState::Initializing(InitializingContractState {
-                candidates: Candidates { candidates },
-                threshold,
-                pk_votes: PkVotes::new(),
-            }),
-            pending_requests: LookupMap::new(b"m"),
-            request_counter: 0,
+    fn add_request(&mut self, payload: &[u8; 32], signature: &Option<(String, String)>) {
+        if self.request_counter > 8 {
+            env::panic_str("Too many pending requests. Please, try again later.");
         }
+        if !self.pending_requests.contains_key(payload) {
+            self.request_counter += 1;
+        }
+        self.pending_requests.insert(payload, signature);
     }
 
-    // This function can be used to transfer the MPC network to a new contract.
-    #[init]
-    pub fn init_running(
-        epoch: u64,
-        participants: BTreeMap<AccountId, ParticipantInfo>,
-        threshold: usize,
-        public_key: PublicKey,
-    ) -> Self {
-        log!(
-            "init_running: signer={}, epoch={}, participants={}, threshold={}, public_key={:?}",
-            env::signer_account_id(),
-            epoch,
-            serde_json::to_string(&participants).unwrap(),
-            threshold,
-            public_key
-        );
-        MpcContract {
-            protocol_state: ProtocolContractState::Running(RunningContractState {
-                epoch,
-                participants: Participants { participants },
-                threshold,
-                public_key,
-                candidates: Candidates::new(),
-                join_votes: Votes::new(),
-                leave_votes: Votes::new(),
-            }),
-            pending_requests: LookupMap::new(b"m"),
-            request_counter: 0,
-        }
+    fn remove_request(&mut self, payload: &[u8; 32]) {
+        self.pending_requests.remove(payload);
+        self.request_counter -= 1;
     }
 
-    pub fn state(self) -> ProtocolContractState {
-        self.protocol_state
+    pub fn respond(&mut self, payload: [u8; 32], big_r: String, s: String) {
+        if let ProtocolContractState::Running(state) = &self.protocol_state {
+            let signer = env::signer_account_id();
+            if state.participants.contains_key(&signer) {
+                log!(
+                    "respond: signer={}, payload={:?} big_r={} s={}",
+                    signer,
+                    payload,
+                    big_r,
+                    s
+                );
+                if self.pending_requests.contains_key(&payload) {
+                    self.pending_requests.insert(&payload, &Some((big_r, s)));
+                }
+            } else {
+                env::panic_str("only participants can respond");
+            }
+        } else {
+            env::panic_str("protocol is not in a running state");
+        }
     }
 
     pub fn join(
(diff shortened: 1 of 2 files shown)


contract/src/lib.rs
```diff
@@ -47,15 +46,49 @@ pub enum ProtocolContractState {
...
+#[derive(BorshDeserialize, BorshSerialize)]
+pub enum VersionedMpcContract {
+    V0(MpcContract),
+}
+
+impl Default for VersionedMpcContract {
+    fn default() -> Self {
+        VersionedMpcContract::V0(MpcContract::default())
+    }
+}
```
ChaoticTempest, line 58: hmm, why do we have a `Default` impl? `MpcContract::default()` call will panic because we made it so with the `PanicOnDefault` attribute. We shouldn't have a default impl for `VersionedMpcContract` then since that would give away to calling into an `init` function and overwrite contract state without our control

contract/src/lib.rs
```diff
@@ -417,42 +498,31 @@ impl MpcContract {
...
 
-    fn remove_request(&mut self, payload: &[u8; 32]) {
-        self.pending_requests.remove(payload);
-        self.request_counter -= 1;
+        Self::V0(MpcContract {
+            protocol_state: old_contract.protocol_state,
+            pending_requests: LookupMap::new(b"m"),
+            request_counter: 0,
+        })
     }
```
ChaoticTempest, line 527: we need a test such that any state changes gets reflected into this `migrate_state` function as well. So something like a unit test that calls into this migrate_state function

contract/src/lib.rs
```diff
@@ -327,29 +379,33 @@ impl MpcContract {
...
-            payload,
-            path,
-            key_version
-        );
-        match self.pending_requests.get(&payload) {
-            None => {
-                self.add_request(&payload, &None);
-                log!(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
-                Self::ext(env::current_account_id()).sign_helper(payload, 0)
+    fn get_signature_per_payload(&self, payload: [u8; 32]) -> Option<Option<(String, String)>> {
```
ChaoticTempest, line 382: nit(rust): we should not have functions with `get_` prefix unless it's `get()` or `get_mut()` functions. So we should have `signature_per_payload` instead

ChaoticTempest (commented):
overall looks good to me besides the points I mentioned

contract/src/lib.rs
ppca: We need to have Default trait implemented because `#[near_bindgen]` requires Default trait. The example [here](https://github.com/near/near-sdk-rs/blob/9aa5eb0b7865e76ac7e89b7cbf46e268be4b56e2/examples/versioned/src/lib.rs) uses #[near(contract_state)] which is not included in the version of near-sdk we are using.
I could make the default function directly panic instead.  `PanicOnDefault` does not work for enums

contract/src/lib.rs
ppca: added

contract/src/lib.rs
ChaoticTempest: I guess the `PanicOnDefault` was not considered very well for enums. Let's just have our `Default` impl panic then

contract/src/lib.rs
```diff
@@ -49,15 +48,69 @@ pub enum ProtocolContractState {
...
+        MpcContract {
+            protocol_state: ProtocolContractState::Initializing(InitializingContractState {
+                candidates: Candidates::default(),
+                threshold: 2,
+                pk_votes: PkVotes::new(),
+            }),
+            pending_requests: LookupMap::new(b"m"),
+            request_counter: 2,
+        }
+    }
```
ChaoticTempest, line 109: This should not be apart of the contract itself since that will bloat the code size and increase the storage staking cost of deploying this contract

ChaoticTempest (approved):

contract/src/lib.rs
```diff
@@ -418,69 +516,46 @@ impl MpcContract {
...
-            self.pending_requests.remove(payload);
+        match self {
+            Self::V0(mpc_contract) => {
+                mpc_contract.clean_payloads(payloads, counter);
+            }
         }
-        self.request_counter = counter;
+    }
+
+    /// This is the root public key combined from all the public keys of the participants.
```
volovyks, line 542: We will need to refactor the contract. It is getting too big. But for now, let's keep public user-facing functions on top.

contract/tests/tests.rs
```diff
@@ -37,3 +38,19 @@ async fn test_contract_can_not_be_reinitialized() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[test]
+fn test_old_state_can_be_migrated_to_v0() -> anyhow::Result<()> {
+    let old_contract = MpcContract::init(3, BTreeMap::new());
+    env::state_write(&old_contract);
```
volovyks, line 45: This code is not executed in Wasmer. Where state is written to?
We need to write an integration test using workspaces to test this logic. And to test it properly we will need to have actual old and new states.

contract/src/lib.rs
```diff
@@ -418,69 +516,46 @@ impl MpcContract {
...
+    pub fn public_key(self) -> PublicKey {
```
volovyks, line 543: We should proxy all user-facing functions. All the tmp and service functions are not that important.

contract/src/lib.rs
volovyks: Should we separate impl into 2 or 3 blocks? Like: user-facing functions, interactions with node and init, helper functions. cc @ChaoticTempest

contract/src/lib.rs
ppca: what do you mean by proxy all user-facing functions?

contract/tests/tests.rs
volovyks: We can have a dummy structure representing an old state V0, then V1 will be the current one. I don't think that is necessary now. We will add such a test on the next state change.

contract/src/lib.rs
volovyks: We care about breaking changes for user and node-facing functions. That is why we should wrap all of them. All the service functions can have breaking changes.

contract/src/lib.rs
ChaoticTempest: Yeah, that would be ideal to keep public facing functions at the top of a `lib.rs` with a `private.rs` to do the helper functions. But most of this is just some house keeping. For the purpose of having things visible as more people come to read the contract, the fastest thing to do right now is just surface the most important public functions to the very top without much refactoring into separate files. We can do this in a separate PR

contract/tests/tests.rs
ChaoticTempest: The `env::state_write` writes to a `MockBlockchain` in the case of a unit test. We don't need to go into workspaces for this since all this is testing is the serialization and that's what we want to test directly. 

> We can have a dummy structure representing an old state V0, then V1 will be the current one. I don't think that is necessary now. We will add such a test on the next state change.

yeah, we need to maintain a list of all structs that we used before in these tests

ChaoticTempest (dismissed):

ChaoticTempest (approved):

contract/src/lib.rs
```diff
@@ -49,23 +48,55 @@ pub enum ProtocolContractState {
 }
 
 #[near_bindgen]
-#[derive(BorshDeserialize, BorshSerialize, PanicOnDefault)]
+#[derive(BorshDeserialize, BorshSerialize, Debug)]
+pub enum VersionedMpcContract {
+    V0(MpcContract),
```
mikedotexe, line 53: +1

PR #560, opened 2024-04-03
fix: stop infinite process of presignature message missing triples

ppca:
Currently, we would push presignature message to leftover messages if not both triples are found. This means we will likely process these messages over and over and never get rid of them. 

This change will time such message out if > 1 min (presignature timeout threshold), and this message won't be retained in the queue.


volovyks (commented):
I've checked the logs and there are no messages (`no new messages received`) without signature requests. There is no infinite loop of messages, but maybe they stay in the loops for a long time.

We should choose one approach here and use it in triple, presignature, and signature generation protocols since the problematics is the same.

ppca:
> I've checked the logs and there are no messages (`no new messages received`) without signature requests. There is no infinite loop of messages, but maybe they stay in the loops for a long time.
> 
> We should choose one approach here and use it in triple, presignature, and signature generation protocols since the problematics is the same.

The message is `could not initiate non-introduced presignature: triple might not have completed for this node yet` or `TripleIsMissing`. I think this happens when a presiganture generation involves triples that this node does not have.

ChaoticTempest (commented):
So for the most part, we're seeing these logs when we're waiting for triples to get generated, and mostly occurs when we don't have a stockpile of triples yet, when the presignature manager is waiting for specific triples. So we should be distinguishing between `TripleIsMissing` and `TripleIsGenerating` sort of errors

ppca:
> So for the most part, we're seeing these logs when we're waiting for triples to get generated, and mostly occurs when we don't have a stockpile of triples yet, when the presignature manager is waiting for specific triples. So we should be distinguishing between `TripleIsMissing` and `TripleIsGenerating` sort of errors

The case that's more problematic is when a node does not have a triple, and not going to generate it (for example, if somehow datastore did not persist triples successfully or the node goes offline at an unfortunate time), but it will always be processing the msg that asks the node to generate a presignature with that triple. Such msg will stay in the queue forever.

TripleIsGenerating makes sense. I'm thinking if the triple is in triple_manager.generators, we throw TripleIsGenerating and push the presignature msg into the leftover messages; if the triple is not in generators, we throw TripleIsMissing and time it out using presignature timeout.

ppca:
> We should choose one approach here and use it in triple, presignature, and signature generation protocols since the problematics is the same.

Agree we should use the same approach.

ppca:
removed requested altogether. 
different between TripleIsGenerating and TripleIsMissing, and skip presignature and signature messages that were sent more than 1 min ago.

node/src/protocol/presignature.rs
```diff
@@ -409,6 +427,7 @@ impl PresignatureManager {
                             epoch: self.epoch,
                             from: self.me,
                             data,
+                            timestamp: Utc::now().timestamp() as u64
```
volovyks, line 430: No need to convert if we will use elapsed()

node/src/util.rs
```diff
@@ -83,3 +84,16 @@ pub fn get_triple_timeout() -> Duration {
         .unwrap_or_default()
         .unwrap_or(crate::types::PROTOCOL_TRIPLE_TIMEOUT)
 }
+
+pub fn is_elapsed_longer_than_timeout(timestamp_sec: u64, timeout: Duration) -> bool {
```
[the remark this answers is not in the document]
ppca: that elapsed works with Instant, which records the relative timestamp in the system, not the absolute timestamp. Since we are sending the timestamp as a field in the messages, it needs to be comprehensible in all nodes' system, thus using u64 timestamp to represent the time.

node/src/util.rs
```diff
@@ -83,3 +84,16 @@ pub fn get_triple_timeout() -> Duration {
...
+    if let LocalResult::Single(msg_timestamp) = Utc.timestamp_opt(timestamp_sec as i64, 0) {
+        let now_datetime: DateTime<Utc> = Utc::now();
+        // Calculate the difference in seconds
+        let elapsed_duration = now_datetime.signed_duration_since(msg_timestamp);
+        let timeout = chrono::Duration::seconds(timeout.as_secs() as i64)
+            + chrono::Duration::nanoseconds(timeout.subsec_nanos() as i64);
+        elapsed_duration > timeout
+    } else {
+        false
```
volovyks, line 97: If we return false in case of failure, such message will never be deleted from the queue.

ppca:
hmmm I realized adding the timestamp as required to messages may make it non-backward compatible. It should be an optional field instead. Will change.

ppca:
> @ppca currently we are updating all the nodes at the same time. We better do it as required now than dealing with Optional later.

Problem is partner testnet is not....they update within an hour but not at the same time

volovyks:
Ok, but let's delete Optional after the update.

node/src/util.rs
ChaoticTempest: Seems like the case where it's not `Single` only happens in the [following](https://docs.rs/chrono/latest/chrono/trait.TimeZone.html#method.timestamp_opt)
> Returns MappedLocalTime::None on out-of-range number of seconds and/or invalid nanosecond, otherwise always returns MappedLocalTime::Single.

So we should be fine for the most part since our timestamp should never be out of range since we created it with `Utc::now().timestamp()` in the first place, but maybe we should add some guardrails to the type itself later when we get the chance, just in case someone ends up setting `timestamp` to be something else not from the `chrono` crate

ChaoticTempest (approved):

PR #565, opened 2024-04-11
feat: separate sign key for signing protocol messages

ChaoticTempest:
This introduces a separate signing key for messages to be sent over the wire for protocols like triple, presignature and signature generation. @ppca this also adds in `multichain-sign-sk-*` to terraform so this introduces a new key needed to be added by partner nodes. This key can be generated via: 
```rs

    #[test]
    fn test() {
        let sk = near_crypto::SecretKey::from_random(near_crypto::KeyType::ED25519);
        println!("{sk}");
    }
```

### DIFF
--- a/infra/multichain-vm-testnet-deprecated/main.tf
+++ b/infra/multichain-vm-testnet-deprecated/main.tf
@@ -36,6 +36,10 @@ module "gce-container" {
         value = data.google_secret_manager_secret_version.cipher_sk_secret_id[count.index].secret_data
       },
       {
+        name  = "MPC_RECOVERY_SIGN_SK"
+        value = data.google_secret_manager_secret_version.sign_sk_secret_id[count.index].secret_data
+      },
+      {
         name  = "AWS_ACCESS_KEY_ID"
         value = data.google_secret_manager_secret_version.aws_access_key_secret_id.secret_data
       },
--- a/integration-tests/src/multichain/containers.rs
+++ b/integration-tests/src/multichain/containers.rs
@@ -42,6 +42,8 @@ impl<'a> Node<'a> {
     ) -> anyhow::Result<Node<'a>> {
         tracing::info!("running node container, account_id={}", account_id);
         let (cipher_sk, cipher_pk) = hpke::generate();
+        let sign_sk =
+            near_crypto::SecretKey::from_seed(near_crypto::KeyType::ED25519, "integration-test");
         let storage_options = ctx.storage_options.clone();
         let args = mpc_recovery_node::cli::Cli::Start {
             near_rpc: ctx.lake_indexer.rpc_host_address.clone(),
@@ -51,6 +53,7 @@ impl<'a> Node<'a> {
             web_port: Self::CONTAINER_PORT,
             cipher_pk: hex::encode(cipher_pk.to_bytes()),
             cipher_sk: hex::encode(cipher_sk.to_bytes()),
+            sign_sk: sign_sk.to_string(),
             indexer_options: mpc_recovery_node::indexer::Options {
                 s3_bucket: ctx.localstack.s3_bucket.clone(),
                 s3_region: ctx.localstack.s3_region.clone(),
--- a/node/src/mesh/mod.rs
+++ b/node/src/mesh/mod.rs
@@ -1,7 +1,24 @@
+use mpc_keys::hpke;
+
 use crate::protocol::contract::primitives::Participants;
 use crate::protocol::ProtocolState;
 
 pub mod connection;
+
+#[derive(Clone, Debug)]
+pub struct NetworkConfig {
+    pub sign_sk: near_crypto::SecretKey,
+    pub cipher_pk: hpke::PublicKey,
+}
+
+impl NetworkConfig {
+    pub fn test() -> Self {
+        Self {
+            sign_sk: near_crypto::SecretKey::from_seed(near_crypto::KeyType::ED25519, "test"),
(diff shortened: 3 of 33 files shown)


ppca:
LGTM, but can we add a script to infra/scripts that partners can run easily so I could point them to it?

ppca:
when we decrypt a message we are fetching the sign_pk from contract state([code](https://github.com/near/mpc-recovery/blob/96ec17f3e78ef31410e50c1a1e3c793cbd51da0a/node/src/protocol/message.rs#L426)), so we need to change contract state's sign_pk's to reflect this. 
How do we plan to change contract state? Can we add code for it? 
Also we'd expect down time before all partners change their node and we push the change to the contract.

ChaoticTempest:
@ppca so how I imagine this would work is that the partners would actually still use their account sk in the mean time, while newer nodes that join the network will have their separate sign secret key. This way we have minimal changes for now, and there's no restriction anyways where a node cannot use their account secret key to sign. This just allows the ability for a separate sign secret key to be provided. The partner network is testnet anyways, so it'll be fine.

> Can we add a script to infra/scripts that partners can run easily so I could point them to it?

Was looking for an easy way to do this, but doesn't seem quite feasible in the format that our keys are required to be in. We might have to have a separate rust project where nodes can compile and run to generate all these keys

node/src/mesh/mod.rs
```diff
@@ -1,8 +1,25 @@
...
 pub mod connection;
 
+#[derive(Clone, Debug)]
+pub struct NetworkConfig {
+    pub sign_sk: near_crypto::SecretKey,
+    pub cipher_pk: hpke::PublicKey,
+}
+
+impl NetworkConfig {
+    pub fn test() -> Self {
```
volovyks, line 15: Does this mean that all the nodes in our tests will have the same keys? What if we break our contract or node logic and mix all the keys? Tests will not catch that, but our real env will be broken after the release.

integration-tests/src/multichain/containers.rs
```diff
@@ -42,6 +42,8 @@ impl<'a> Node<'a> {
     ) -> anyhow::Result<Node<'a>> {
         tracing::info!("running node container, account_id={}", account_id);
         let (cipher_sk, cipher_pk) = hpke::generate();
+        let sign_sk =
```
volovyks, line 45: Why do we need containers again? Should we get rid of it?

volovyks:
@ChaoticTempest @ppca we definitely can use NEAR CLI for ED25519 key generation. Why should we ask partners to do 
that with tests?

volovyks:
@ChaoticTempest @ppca
I agree that current partners can use the same key for signing messages and interactions with NEAR Blockchain. Ideally, they can use the same key for signing messages, but rotate and update their NEAR Key. But that is not necessary, since it's still on testnet. MAinnet partners will have 2 separate keys anyway.

infra/multichain-vm-testnet-deprecated/main.tf
```diff
@@ -35,6 +35,10 @@ module "gce-container" {
         name  = "MPC_RECOVERY_CIPHER_SK"
         value = data.google_secret_manager_secret_version.cipher_sk_secret_id[count.index].secret_data
       },
+      {
```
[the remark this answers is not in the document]
ppca: This means the 3-node internal testnet

ppca:
So for testnet, existing partners can run with what they have, no need to add new sign key, while new joining partners will generate a separate sign key. For mainnet, we will require that all partners start off with separate sign key.

This plan sounds good to me, but then the PR actually requires the MPC_RECOVERY_SIGN_SK to be present once we push the code changes to testnet partners. We could ask partners to do `gcloud compute instances update-container multichain-dev-0 --container-env MPC_RECOVERY_SIGN_SK={near_private_key}` to add the env var to their node.

btw, do we still need account_sk then?

ppca:
> @ChaoticTempest @ppca we definitely can use NEAR CLI for ED25519 key generation. Why should we ask partners to do that with tests?

That will work for me as long as the near cli command can output the exact command we need.

infra/multichain-vm-testnet-deprecated/main.tf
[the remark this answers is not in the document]
volovyks: Seams like our 8 node setup is stable enough, should we get rid of the old 3 node env?

volovyks:
> btw, do we still need account_sk then?

Not sure what you mean. Each node still needs to interact with NEAR Blockchain, so the key is still required.

node/src/mesh/mod.rs
ChaoticTempest: this is actually unused code -- will delete the function. The tests use different keys per node so nothing to worry about

integration-tests/src/multichain/containers.rs
ChaoticTempest: containers is there if we want to run docker tests. It's not bad to have, but we can evaluate this later

infra/multichain-vm-testnet-deprecated/main.tf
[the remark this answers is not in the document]
ppca: yeah we can do that

ChaoticTempest:
@ppca made changes to this PR so that the terraform should only check if the resource is present then set the env for it, which thereby defaults to using the account secret key if it's the sign secret key is not supplied. Can you also check if my terraform code is correct here?

ppca (approved):
LGTM. I'll try re-running the terraform commands after this merges.

ppca (approved):

PR #586, opened 2024-05-02
Make sure when we fail in the busy loop we cleanup

DavidM-D:
We got reports of users getting responses to sign requests on testnet of "Too many pending requests. Please, try again later." when the signing contract was under no load.

I believe this was caused by someone sending a message that failed, the contract cleaning up the request but then immediately panicking. This caused the state cleanup to roll back and the system keeping a limit on the total number of requests to eventually run out of space.

This PR fixes that by panicking on a seperate call and allowing the cleanup to persist.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -384,21 +384,41 @@ impl MpcContract {
                     PromiseOrValue::Value(signature)
                 }
                 None => {
-                    if depth > 30 {
-                        self.remove_request(&payload);
-                        env::panic_str("Signature was not provided in time. Please, try again.");
+                    // Make sure we have enough gas left to do 1 more call and clean up afterwards
+                    // Observationally 30 calls < 300 TGas so 2 calls < 20 TGas
+                    if depth > 30 && env::prepaid_gas() > Gas::from_tgas(20) {
+                        self.fail(
+                            "Signature was not provided in time. Please, try again.",
+                            payload,
+                        )
+                    } else {
+                        log!(&format!(
+                            "sign_helper: signature not ready yet (depth={})",
+                            depth
+                        ));
+                        let account_id = env::current_account_id();
+                        PromiseOrValue::Promise(
+                            Self::ext(account_id).sign_helper(payload, depth + 1),
+                        )
                     }
-                    log!(&format!(
-                        "sign_helper: signature not ready yet (depth={})",
-                        depth
-                    ));
-                    let account_id = env::current_account_id();
-                    PromiseOrValue::Promise(Self::ext(account_id).sign_helper(payload, depth + 1))
                 }
             }
         } else {
-            env::panic_str("unexpected request");
+            self.fail("unexpected request", payload)
         }
+    }
+
+    /// Cleanup the state and fail
+    /// This allows us to return a panic, without rolling back the state from this call
+    pub fn fail(&mut self, message: &str, payload: [u8; 32]) -> PromiseOrValue<(String, String)> {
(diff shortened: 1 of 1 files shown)


contract/src/lib.rs
```diff
@@ -384,23 +384,43 @@ impl MpcContract {
                     PromiseOrValue::Value(signature)
                 }
                 None => {
-                    if depth > 30 {
-                        self.remove_request(&payload);
-                        env::panic_str("Signature was not provided in time. Please, try again.");
+                    // Make sure we have enough gas left to do 1 more call and clean up afterwards
+                    // Observationally 30 calls < 300 TGas so 2 calls < 20 TGas
+                    if depth > 30 && env::prepaid_gas() > Gas::from_tgas(20) {
```
gagdiez, line 389: Since anyway you will fail at 30, I would find the following easier to read (note: pseudo code bellow):

```rust
if depth == 30 {
  self.remove_request(&payload);
  let self_id = env::current_account_id();
  PromiseOrValue::Promise(Self::ext(self_id).fail_helper("Signature was not provided in time. Please, try again.")
 }

```

gagdiez: to be honest, I think you can leave the `> 30` and not need the gas check, since otherwise you would not remove the request (it would go through the else, and call itself again)

otherwise we can change the `>30` to `>29`, and keep old assumptions that there will be gas for one more call

DavidM-D: Probably the more sensible thing is for me to put a test in there checking that the contract can actually manage 30 calls as we change the code. 

What this defends against is us changing the code making it a little more expensive then having our contract not have enough gas to cleanup the state. That being said this is a hotfix, so I can just remove it for now and add it back in, in a more principled way later.

gagdiez: my point is that `env::prepaid_gas() > Gas::from_tgas(20)` has two outcomes:

- either the prepaid gas is enough (i.e. more than 20Tgas) and you try to remove the payload
- the prepaid gas is not enough (i.e. less than 20Tgas) and you try to call yourself again

in both cases you make a cross-contract call, in only one you remove the payload. Except if you were trying to check for remaining gas only:

- if we have less than 20Tgas, then try to remove the payload

In which case you could just check for the gas, and not the depth

P.S. this means that we will now always remove the payload if we didn't manage to sign it in one transaction?

DavidM-D: Fair point, I'll remove the gas check for now then

DavidM-D:
The audit is failing because of an updated rust version, but I'm wiling to merge it for now and fix that in a subsequent PR because this is blocking people from using chain signatures on testnet

gagdiez (approved):

contract/src/lib.rs
```diff
@@ -384,23 +384,44 @@ impl MpcContract {
...
-                        "sign_helper: signature not ready yet (depth={})",
-                        depth
-                    ));
-                    let account_id = env::current_account_id();
-                    PromiseOrValue::Promise(Self::ext(account_id).sign_helper(payload, depth + 1))
                 }
             }
         } else {
-            env::panic_str("unexpected request");
+            self.fail("unexpected request", payload)
```
gagdiez, line 408: this could also be left as before, since the payload will not exist, and removing it will most probably cause an error

DavidM-D: Right you are and now fail is only used once so we can inline it.

gagdiez (approved):

PR #588, opened 2024-05-08
fix: only retry signature generators with proposer = me

ppca:
check if the failed generator's proposer == me, only take my presignature and retry if it's true.


DavidM-D:
Feel free to merge it with cargo audit complaining, I'll fix it in a separate PR

PR #589, opened 2024-05-08
Lowered the minimum gas limit and upped the depth

DavidM-D:
Observationally 30 transactions uses about 160 TG of Gas, so 250 should be plenty. The protocol team says that this should come down soon.

The current gas limit makes it impossible for people to call the sign method using another smart contract.

In the future we should abort when the remaining gas in the call starts to get low to avoid the risk of an out of gas panic causing the system to not clean up after itself.

### DIFF
--- a/contract/src/lib.rs
+++ b/contract/src/lib.rs
@@ -9,7 +9,7 @@ use primitives::ParticipantInfo;
 use primitives::{CandidateInfo, Candidates, Participants, PkVotes, Votes};
 use std::collections::{BTreeMap, HashSet};
 
-const GAS_FOR_SIGN_CALL: Gas = Gas::from_gas(3 * 100_000_000_000_000);
+const GAS_FOR_SIGN_CALL: Gas = Gas::from_gas(250_000_000_000_000);
 
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug)]
 pub struct InitializingContractState {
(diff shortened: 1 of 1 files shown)


contract/src/lib.rs
```diff
@@ -387,7 +387,8 @@ impl MpcContract {
                     // Make sure we have enough gas left to do 1 more call and clean up afterwards
                     // Observationally 30 calls < 300 TGas so 2 calls < 20 TGas
                     // We keep one call back so we can cleanup then call panic on the next call
-                    if depth > 29 {
+                    // Start cleaning up if there's less than 25 teragas left regardless of how deep you are.
+                    if depth > 30 || env::prepaid_gas() < Gas::from_tgas(25) {
```
ChaoticTempest, line 391: I feel like we should not have a set depth anymore, since this is purely already limited by the amount of gas a user specifies which is better for consistency. Users should feel free to burn however much gas they want to
```suggestion
                    if env::prepaid_gas() < Gas::from_tgas(25) {
```

ChaoticTempest (approved):
LGTM, but I wonder if we should remove the following restriction. Up to you but looks good to me to merge either ways

contract/src/lib.rs
DavidM-D: Right now we have a nice mapping between the timeout on the contract and the timeout on the MPC network. I'd probably need to rejig the congestion counter so that a bunch of requests didn't time out on the contract but not on the MPC service, or alternatively make the timeout on the MPC network configurable per request based on the gas attached.

DavidM-D: I'll look at that in a subsequent PR.

PR #594, opened 2024-05-14
fix: allow triple memory only

ChaoticTempest:
This allows us to triples from memory only in the case the triples were never saved to datastore somehow and rely purely on local memory. This is good in the case that datastore service is down somehow


ppca (approved):
LGTM

PR #595, opened 2024-05-14
fix: possible deadlock in load triple

ppca:
Issue:
Node state never advanced from Starting to Started. 
I observe the logs where "loading triples using datastore" were printed 3 times, but not seeing "loading triples success" or "triple load failed." message being printed. 

Resolution:
I'm not able to find exactly what went wrong, I know state advancing was stuck at loading triples. I deleted all triples in datastore, then the node state start advancing.

Possible cause:
I looked through the code for `async fn load_triples()`, I'm not able to figure out why "loading triples using datastore" were printed 3 times, but not seeing "loading triples success" or "triple load failed." message being printed. 
I know we had some schema incompatibility issue, and possible deadlocking, but I'm not sure if they indicate no success or failure messages.

this PR:
address the potential deadlock issue

### DIFF
--- a/node/src/protocol/consensus.rs
+++ b/node/src/protocol/consensus.rs
@@ -704,14 +704,13 @@ async fn load_triples<C: ConsensusCtx + Send + Sync>(
     ctx: C,
 ) -> Result<Vec<TripleData>, ConsensusError> {
     let triple_storage = ctx.triple_storage();
-    let read_lock = triple_storage.read().await;
     let mut retries = 3;
     let mut error = None;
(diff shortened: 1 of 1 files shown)


ChaoticTempest (approved):
weird, I wonder if the datastore service when it's fetching entries is creating a separate task which leads to the main retry while loop being able to call multiple `triple_storage.load` calls and each conflicting with each other. By moving the lock into the loop, the loop wouldn't be able to advance to the next `load` call and thus not have any sort of conflicts

PR #597, opened 2024-05-15
chore: Made failed generators only get added on self proposers

ChaoticTempest:
This cleanups the signature handling we have in the crypto loop to be inside the `SignatureManager` since that function was getting hairy. Also, we did not need to have failed generators be added every single time to be handled, so we only need to directly keep around the failed ones that we originally introduced into the system.

Also created https://github.com/near/mpc-recovery/issues/596 for our future handling of priority of sign requests

### DIFF
--- a/node/src/protocol/signature.rs
+++ b/node/src/protocol/signature.rs
@@ -244,13 +244,13 @@ impl SignatureManager {
         ))
     }
 
-    pub fn retry_failed_generation(
+    fn retry_failed_generation(
         &mut self,
         receipt_id: CryptoHash,
         failed_generator: &FailedGenerator,
         presignature: Presignature,
         participants: &Participants,
-    ) -> Option<()> {
+    ) -> Result<(), InitializationError> {
         tracing::info!(receipt_id = %receipt_id, participants = ?participants.keys_vec(), "restarting failed protocol to generate signature");
         let generator = Self::generate_internal(
             participants,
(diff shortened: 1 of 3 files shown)


node/src/protocol/signature.rs
```diff
@@ -439,6 +437,79 @@ impl SignatureManager {
...
+                );
+                failed_presigs.push(presignature);
+                continue;
+            }
+
+            alternate = !alternate;
+            // NOTE: this prioritizes new requests first then alternates to reattempt failed ones.
+            // TODO: we need to decide how to prioritize certain requests over others such as with gas or
+            // time of when the request made it into the NEAR network. Ticket: https://github.com/near/mpc-recovery/issues/596
+            if alternate {
```
ppca, line 470: if we got no requests in my_requests, but some in failed_generators, since the handle_request fn will always start with alternate = true when entering this if logic, then none of the failed_generators will ever get processed.

If we hope to process only one request/failed_generator on each handle_requests fn call (like we previously had in code, we can do continue when `my_requests.keys().next().cloned()` or `my_requests.remove(&receipt_id)` return None, and break after self.generate().

ChaoticTempest: ahh, I think the `break` just ruins this setup. I just got rid of it entirely, but I made it handle failed ones first for now

ppca (approved):
LGTM

volovyks (approved):

PR #600, opened 2024-05-20
remove infra code for non-vm

ppca:
@kmaus-near Let's also terraform destroy all these before we merge this.


volovyks (approved):
Not sure why the tests are failing.

kmaus-near (approved):

PR #607, opened 2024-05-22
fix terraform code for getting sign_sk

ppca:


volovyks (approved):
LGTM
let's change it back after it is released on all current envs

PR #611, opened 2024-05-23
Signature deposit

volovyks:
We will need to add a [refund](https://github.com/near/near-sdk-rs/blob/5a9acaedc95c5721d2088f263bc99e3de574decf/near-contract-standards/src/non_fungible_token/utils.rs#L30) functionality after wee implement yield/resume approach.


volovyks:
@ChaoticTempest we do not have Relayer setup in multichain test and I do not think we need it just to test if deposits work with relayer. But I will doublecheck if it works before we merge this.

volovyks:
I had a call with @r-near. We should be fine with passing DelagateActions with deposits.

contract/src/lib.rs
```diff
@@ -160,6 +172,23 @@ impl VersionedMpcContract {
         }
     }
 
+    /// To avoid overloading the network with too many requests,
+    /// we ask for a small deposit for each signature request.
+    /// The fee changes based on how busy the network is.
+    pub fn signature_deposit(&self) -> u128 {
```
ChaoticTempest, line 178: now that I think about this a bit more, I'm not sure if allowing people to call this function to check what's the required deposit would be great experience, since this amount can change in the time they call this function and when they call `sign`. This would cause a lot of retries which could further congest the network.

So instead, to avoid having this situation, we should ask the users to always make the highest deposit we can take when calling into `sign`, and then we can refund the difference to the account if the cheaper requests are possible

volovyks: People may want to spend 1 yN maximum.
If they provide more than we need, we will definitely refund (separate PR).
You may be right that it should not be a part of the API since it will work in ~90% of cases which is not enough. People can call `sign()` with the maximum amount they are ready to spend and get an error if the network is under load.

ailisp: Yeah I'm also think people would always deposit the maximum deposit to ensure it get processed. And we can refund the difference based on load

volovyks: we will add refund logic after yield/resume is merged

ChaoticTempest (approved):

ailisp (dismissed):

ChaoticTempest (approved):

PR #612, opened 2024-05-24
fix: add port to my_address so join method works

ppca:


volovyks (dismissed):

node/src/cli.rs
```diff
@@ -221,10 +221,15 @@ pub fn run(cmd: Cli) -> anyhow::Result<()> {
                     ));
 
                     let sign_sk = sign_sk.unwrap_or_else(|| account_sk.clone());
-                    let my_address = my_address.unwrap_or_else(|| {
-                        let my_ip = local_ip().unwrap();
-                        Url::parse(&format!("http://{my_ip}:{web_port}")).unwrap()
-                    });
+                    let my_address = my_address
+                        .map(|mut addr| {
+                            let _ = addr.set_port(Some(web_port));
```
ChaoticTempest, line 226: we should unwrap and fail early just in case the `addr` was inputted incorrectly
```suggestion
                            addr.set_port(Some(web_port)).unwrap();
```

ChaoticTempest (approved):

PR #614, opened 2024-05-25
chore(refactor): restructure project + upgrade dependencies/transitive-dependencies

ChaoticTempest:
This upgrades all our near related dependencies to `0.21` to resolve security vulnerabilities. The project was restructured to facilitate this due to the fastauth project being in conflict with these dependencies. This is how our folder structure looks like now:

```
integration-tests/
- chain-signatures
- fastauth
mpc-recovery/
node/
contract/
keys/
...
```

Further restructuring will happen after this PR merges due to potential merge conflicts:
```
integration-tests/
- chain-signatures/
- fastauth/
chain-signatures/
- node/
- contract/
- keys/
fastauth
...
```

**Note** that we no longer have a top-level `Cargo.toml`, which means we can no longer use the `cargo` command at the root of the project. We'll need to `cd` directly into the folder to use the command, ~~and also watch out potentially for rust-analyzer since it won't be watching our project anymore at the root.~~

This also reduces our build times now since we have two distinct projects when going to run tests instead of compiling everything in `integration-tests`.

Nix might also be failing since I didn't test that 🤷


ChaoticTempest:
nevermind about rust-analyzer not working in root folder. Got it to work by including `.vscode/settings.json` entries

ChaoticTempest:
Audits will also continue to fail since I haven't completely upgraded everything yet. That will be left for the next PR

node/src/protocol/consensus.rs
```diff
@@ -636,24 +641,20 @@ impl ConsensusProtocol for JoiningState {
...
+                                "url": ctx.my_address(),
+                                "cipher_pk": ctx.cfg().network_cfg.cipher_pk.to_bytes(),
+                                "sign_pk": ctx.cfg().network_cfg.sign_sk.public_key(),
+                            }))
+                            .max_gas()
+                            .transact()
                             .await
-                            .unwrap();
+                            .map_err(|err| {
+                                tracing::error!(?err, "failed to join the participant set");
```
ppca, line 655: nice.

.github/workflows/multichain-integration.yml
```diff
@@ -43,7 +44,7 @@ jobs:
       - name: Pull Relayer & Sandbox Docker Images
         run: |
           docker pull ghcr.io/near/os-relayer:12ba6e35690df3979fce0b36a41d0ca0db9c0ab4
-          docker pull ghcr.io/near/near-lake-indexer:e6519c922435f3d18b5f2ddac5d1ec171ef4dd6b
+          docker pull ghcr.io/near/near-lake-indexer:node-1.38
```
ppca, line 47: why node-1.38?

ppca (approved):

.github/workflows/multichain-integration.yml
ChaoticTempest: because node-1.39 doesn't work well with `near-crypto 0.21`. I'll update to `near-crypto 0.22` afterwards which does. Just didn't want to add more complexity to this PR

volovyks (approved):
Great work! Let's merge it ASAP.

PR #615, opened 2024-05-26
fix: store reshared key after generation

ppca:
We did not store the new key share to secret manager previously, so whenever restart, we don't have the correct key share for the epoch anymore


ChaoticTempest:
wow, nice catch on this. Can we also add in a test (or maybe add it to already existing reshare test) to check on whether or not after resharing, we were able to store this secret value?

ppca:
> wow, nice catch on this. Can we also add in a test (or maybe add it to already existing reshare test) to check on whether or not after resharing, we were able to store this secret value?

Yep we can do that. Thanks to now we have the local disk storage secret storage option. :D

ChaoticTempest (approved):
gonna merge this one for now. @ppca I'll let you add the test when you get the chance

PR #621, opened 2024-05-30
Signed responses

DavidM-D:
Make the smart contract verify that the responses we're getting back from the nodes are valid.

A large part of the change is move dependencies into crypto_shared so it can be used in the contract for verification.

This has a breaking change in that payloads are no longer little endian and are big endian (standard in our industry).

A subsequent PR will:
- Get the native NEAR verification function working
- Expose the well typed API
- Unreverse the epsilon value
- Force validation of scalars (not added in this PR)
- Abstract the "Serializable" trait from its concrete types


CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
+
+## 0.3.0
+
+- Paylaod hash scalars are now big endian. In general this means that clients of the contract should no longer reverse their hashed payload before sending it to the MPC contract.
```
ailisp, line 5:
```suggestion
- Payload hash scalars are now big endian. In general this means that clients of the contract should no longer reverse their hashed payload before sending it to the MPC contract.
```

CHANGELOG.md
```diff
@@ -0,0 +1,6 @@
+# Changelog
+
+## 0.3.0
```
ailisp, line 3: Also update version in contract/Cargo.toml?

crypto-shared/src/kdf.rs
```diff
@@ -0,0 +1,89 @@
...
+    // TODO: Use a key derivation library instead of doing this manually.
+    // https://crates.io/crates/hkdf might be a good option?
+    //
+    // ',' is ACCOUNT_DATA_SEPARATOR from nearcore that indicate the end
+    // of the accound id in the trie key. We reuse the same constant to
+    // indicate the end of the account id in derivation path.
+    let derivation_path = format!("{EPSILON_DERIVATION_PREFIX}{},{}", predecessor_id, path);
+    let mut hasher = Sha256::new();
+    hasher.update(derivation_path);
+    let mut bytes = hasher.finalize();
```
ailisp, line 25: consider native [near_sdk::env::sha256](https://docs.rs/near-sdk/latest/near_sdk/env/fn.sha256.html) instead of wasm-side sha256 to be more efficient?

crypto-shared/src/kdf.rs
```diff
@@ -0,0 +1,89 @@
...
+    )?
+    .to_encoded_point(false);
+    if public_key == found_pk {
+        return Ok(());
+    }
+
+    anyhow::bail!("cannot use either recovery id (0 or 1) to recover pubic key")
+}
+
+// #[cfg(not(target_arch = "wasm32"))]
```
ailisp, line 67: This and `// #[cfg(target_arch = "wasm32")]` conditional compile seems unnecessary, as near-sdk handles mock of ecrecover etc. on non wasm case (by importing vmlogic from nearcore): https://github.com/near/near-sdk-rs/blob/master/near-sdk/src/environment/mock/mocked_blockchain.rs

DavidM-D: Ah excellent point

crypto-shared/src/kdf.rs
DavidM-D: Will do

integration-tests/chain-signatures/tests/actions/mod.rs
```diff
@@ -168,36 +265,36 @@ pub async fn single_payload_signature_production(
...
 
-    // TODO: get hashed values on the flight
-    let payload_hash: [u8; 32] = [
-        99, 13, 205, 41, 102, 196, 51, 102, 145, 18, 84, 72, 187, 178, 91, 79, 244, 18, 164, 156,
-        115, 45, 178, 200, 171, 193, 184, 88, 27, 215, 16, 221,
-    ];
-    let payload_hash_scalar = k256::Scalar::from_bytes(&payload_hash); // TODO: why do we need both reversed and not reversed versions?
-    let mut payload_hash_reversed: [u8; 32] = payload_hash;
-    payload_hash_reversed.reverse();
+    payload_hash.reverse();
```
volovyks, line 284: We are still reversing the hash here.

volovyks (commented):
Great work @DavidM-D !

contract/Cargo.toml
```diff
@@ -12,6 +12,8 @@ near-sdk = { version = "=5.1.0", features = ["legacy", "unit-testing"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 schemars = "0.8"
+k256 = { version = "0.13.1", features = ["sha256", "ecdsa", "serde", "arithmetic", "expose-field"] }
+crypto-shared = { path = "../crypto-shared" }
```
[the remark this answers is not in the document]
volovyks: I see, you already added those structures.

crypto-shared/src/kdf.rs
DavidM-D: Do you mind if I move this to the next PR? The merge conflicts are killing me

integration-tests/chain-signatures/tests/actions/mod.rs
DavidM-D: This code was and still is a bit of a mess.

Previously converting a Scalar to bytes reversed the bytes and converted to a Scalar. 

The big_r and s values were generated using chain signatures from an older commit, therefore the signature is generated against a reversed hash.

You now have to reverse the bytes to generate the same Scalar and therefore Signature and this tests that. I'll write a comment to this effect, after this change has rolled out we should update this test using new values.

crypto-shared/src/kdf.rs
```diff
@@ -0,0 +1,89 @@
...
+    // https://crates.io/crates/hkdf might be a good option?
+    //
+    // ',' is ACCOUNT_DATA_SEPARATOR from nearcore that indicate the end
+    // of the accound id in the trie key. We reuse the same constant to
+    // indicate the end of the account id in derivation path.
+    let derivation_path = format!("{EPSILON_DERIVATION_PREFIX}{},{}", predecessor_id, path);
+    let mut hasher = Sha256::new();
+    hasher.update(derivation_path);
+    let mut bytes = hasher.finalize();
+    // Due to a previous bug in our Scalar conversion code, this hash was reversed, we reverse it here to preserve compatibility, but will likely change this later.
```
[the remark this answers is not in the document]
DavidM-D: I'm just trying to break up this PR a bit, I'll get it through before the 0.3.0 release

contract/src/lib.rs
```diff
@@ -596,9 +635,15 @@ impl VersionedMpcContract {
         }
     }
 
-    fn sign_result(&self, request: &SignRequest) -> Option<Option<SignResult>> {
+    fn sign_result(&self, request: &SignatureRequest) -> Option<Option<SignatureResponse>> {
         match self {
             Self::V0(mpc_contract) => mpc_contract.pending_requests.get(request),
         }
     }
+
```
volovyks, line 643: Do we need this?

contract/src/lib.rs
```diff
@@ -70,36 +73,53 @@ impl Default for VersionedMpcContract {
     }
 }
 
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone)]
+pub struct SignatureRequest {
```
volovyks, line 77: Why do we need SignatureRequest and SignRequest? Can we use 1 struct?

volovyks (approved):
We can merge this to avoid future conflicts and unblock yield/resume. Great contribution!

contract/src/lib.rs
DavidM-D: SignatureRequest describes all the information about a request required by the signing protocol, SigRequest is the structure of the request sent to the contract, so they're different. That being said, we can probably come up with better names.

PR #624, opened 2024-06-03
near-sdk dependency for contract: move to the commit with yield_resume merged

ppca:
also added the version for crypto-shared so compile is success.
near-sdk commit history: https://github.com/near/near-sdk-rs/commits/master/


ppca:
man I need to resolve merge conflict first.

volovyks (approved):

PR #625, opened 2024-06-03
fix: local build.rs not running on changes

ChaoticTempest:
This fixes local build.rs script not running correctly due to changed folder structure. Also moved `crypto-shared` folder into `chain-signatures` since that would be annoying later to deal with.


integration-tests/chain-signatures/build.rs
```diff
@@ -80,11 +80,22 @@ async fn build_multichain_contract(release: bool) -> anyhow::Result<ExitStatus>
...
+        }
+    }
+}
+
 fn main() -> anyhow::Result<()> {
     println!("cargo:rerun-if-changed=build.rs");
-    println!("cargo:rerun-if-changed=../../contract/");
-    println!("cargo:rerun-if-changed=../../keys/");
-    println!("cargo:rerun-if-changed=../../node/");
+    rerun_directory("../../chain-signatures/");
```
ailisp, line 98: will this result multiple rerun if more than one dir changes?

ailisp (approved):
LGTM!

integration-tests/chain-signatures/build.rs
ChaoticTempest: This build script only gets ran when you build the integration-tests package, so it'll only be ran once. The `rerun_directory` just has it listens to specific timestamps for a list of directories, and if any of them changes, it'll trigger a rebuild the next time we go to build this test package

PR #629, opened 2024-06-05
fix: potential panics when calling into contract

ChaoticTempest:
This fixes a couple `unwrap`s that happen when voting. This can cause some errors to happen when the chain is slow. Also no longer using the near-fetch git repo


ppca (approved):

volovyks (approved):
Looks great!

PR #630, opened 2024-06-05
Optimize ci

ailisp:
Previous cache has a size limit of 10G and it is not sufficient, result in fresh build everytime. This PR uses a unlimited cache to reduce build time. In the best case where dependency crate is not changed and only our code changes, it reduce 20 minutes of build time of unit (28min -> 6min), fastauth integration (33min -> 13min) and chain signature integration tests (35min -> 16min).


volovyks (approved):
That sounds great!
Is it a free solution?

ChaoticTempest (approved):
Very nice, and I didn't know about warp builds until this PR. Looks great to add

volovyks:
https://docs.warpbuild.com/runners
Looks like they have M1 machines. It can be useful for us at some point.

volovyks:
We may need to build M1 Docker images, I'm not sure if it is required to run tests on M1 instances.

ChaoticTempest:
Don't think it would change anything though, since what was broken wasn't that we didn't have M1 docker builds but that colima (the docker virtualization tooling) was flaky on macos and we never bothered to get it working again.

PR #636, opened 2024-06-06
fix: Added retrying on indexer failures

ChaoticTempest:
Noticed that indexer can potentially fail but won't report the error until we go to join the handle much later. This is due to the fact that panics are per thread here and would require getting the panic hook to truly handle it. Which is quite a hassle to deal with for now.

There's one minor thing this doesn't address is the cancellation of threads, which means that the when we go to join the thread handle at the end of the run call, it will potentially not join due to indexer being alive still. We would need to add some cancellation mechanisms such as sending over a message to kill the indexer loop, but that's too much work right now for this simple fix.

### DIFF
--- a/chain-signatures/node/src/cli.rs
+++ b/chain-signatures/node/src/cli.rs
@@ -161,6 +161,35 @@ impl Cli {
     }
 }
 
+fn spinup_indexer(
+    options: &indexer::Options,
+    mpc_contract_id: &AccountId,
+    account_id: &AccountId,
+    sign_queue: &Arc<RwLock<SignQueue>>,
+    gcp: &GcpService,
+) -> std::thread::JoinHandle<()> {
+    let options = options.clone();
+    let mpc_contract_id = mpc_contract_id.clone();
+    let account_id = account_id.clone();
+    let sign_queue = sign_queue.clone();
+    let gcp = gcp.clone();
+    std::thread::spawn(move || {
+        // If indexer fails for whatever reason, let's spin it back up with exponential delay:
+        for i in 0..10 {
+            let options = options.clone();
+            let mpc_contract_id = mpc_contract_id.clone();
+            let account_id = account_id.clone();
+            let sign_queue = sign_queue.clone();
+            let gcp = gcp.clone();
+
+            if let Err(err) = indexer::run(options, mpc_contract_id, account_id, sign_queue, gcp) {
+                tracing::error!(%err, "indexer failed");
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/cli.rs
```diff
@@ -161,6 +161,35 @@ impl Cli {
...
+    std::thread::spawn(move || {
+        // If indexer fails for whatever reason, let's spin it back up with exponential delay:
+        for i in 0..10 {
+            let options = options.clone();
+            let mpc_contract_id = mpc_contract_id.clone();
+            let account_id = account_id.clone();
+            let sign_queue = sign_queue.clone();
+            let gcp = gcp.clone();
+
+            if let Err(err) = indexer::run(options, mpc_contract_id, account_id, sign_queue, gcp) {
```
ppca, line 185: I guess we need to exit the loop if the ::run() is successful?

ChaoticTempest: so, there's no cancellation when it comes to indexer, since it's ran on a separate thread. So that `indexer::run` will never return anything unless it errors out. The reason why none of our stuff stalls is because the integration tests will directly kill the nodes. I'll add in a `break` to make it explicit just in case we do add interrupt handlers in the future, but this is not an issue at all for us

chain-signatures/node/src/cli.rs
```diff
@@ -161,6 +161,40 @@ impl Cli {
...
+
+            // TODO/NOTE: currently indexer does not have any interrupt handlers and will never yield back
+            // as successful. We can add interrupt handlers in the future but this is not important right
+            // now since we managing nodes through integration tests that can kill it or through docker.
+            let Err(err) = indexer::run(options, mpc_contract_id, account_id, sign_queue, gcp)
+            else {
+                break;
+            };
+            tracing::error!(%err, "indexer failed");
+            std::thread::sleep(std::time::Duration::from_secs(2u64.pow(i)));
```
ailisp, line 193: Is sleeping `2**9=512` seconds too long?

ailisp (dismissed):
Looks good to me!

chain-signatures/node/src/cli.rs
ChaoticTempest: yeah that could be too long. Let's just cap it to something like 5 mins instead of 10mins. And instead of 10 loops, we can just loop infinitely

volovyks (approved):

PR #648, opened 2024-06-14
fix: dockerfile multichain

ppca:
Has been broken since we change folder structure. already tested on develop this works.
This will fix the down dev env.

There were 2 bugs with the Dockerfile.multichain:
1) we copied whole mpc folder instead of `chain-signatures/` folder. the issue there is the top level mpc folder has no `Cargo.toml` file, nothing will be built; 
2) after fixing 1), I find that https://github.com/near/mpc/blob/develop/chain-signatures/.cargo/config.toml#L2 actually ask to put the built binary to `../target`, instead of `target`. But later when we do
`COPY --from=builder /usr/src/app/target/release/mpc-recovery-node /usr/local/bin/mpc-recovery-node`
it is copying from `target` instead of `../target`. Which will copy nothing

This can be verified by built artifact size. If you look at artifacts size: [correct](https://console.cloud.google.com/artifacts/docker/pagoda-discovery-platform-dev/us-east1/multichain-public/multichain-dev/sha256:e1c97663fe2de5e6fe3b3e128ef462e6cca44a444c6e5d135615bb2274b7f5c6?project=pagoda-discovery-platform-dev) ones are 60+mb, [incorrect](https://console.cloud.google.com/artifacts/docker/pagoda-discovery-platform-dev/us-east1/multichain-public/multichain-dev/sha256:ea2a84fe2846009f42f9b0e35cad5405a3e3beb48d605fc341fd669603387679?project=pagoda-discovery-platform-dev) ones are only 50+ mb. I think reflects the fact that actually nothing is built.

Also included a dependency fix. Because the dmd/bump-dependencies branch has been deleted, I updated to a commit hash that merged.

### DIFF
--- a/Dockerfile.multichain
+++ b/Dockerfile.multichain
@@ -12,12 +12,9 @@ RUN sed -i 's#mpc-keys = { path = "../keys" }##' Cargo.toml
 RUN sed -i 's#mpc-contract = { path = "../contract" }##' Cargo.toml
 RUN sed -i 's#crypto-shared = { path = "../crypto-shared" }##' Cargo.toml
 RUN cargo build --release
-COPY . .
-RUN sed -i 's#"mpc-recovery",##' Cargo.toml
-RUN sed -i 's#"integration-tests",##' Cargo.toml
-RUN sed -i 's#"load-tests",##' Cargo.toml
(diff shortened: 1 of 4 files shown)


ChaoticTempest (approved):
nice

PR #649, opened 2024-06-17
Cleaned up the response verification

DavidM-D:
This now does native ecdsa key derivation and the epsilon is unreversed

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -557,7 +557,7 @@ impl SignatureManager {
             let status = response.status();
 
             // Check we returned successfully
-            let FinalExecutionStatus::SuccessValue(_) = status else {
+            let FinalExecutionStatus::SuccessValue(payload) = status else {
                 let err = format!("Publish transaction failed, {:?}", response);
                 tracing::error!(err);
                 return Err(near_fetch::Error::RpcReturnedInvalidData(err));
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -13,7 +13,6 @@ use crypto_shared::{derive_key, PublicKey};
 use crypto_shared::{ScalarExt, SerializableScalar};
 use k256::{Scalar, Secp256k1};
 use mpc_contract::SignatureRequest;
-use near_primitives::views::FinalExecutionStatus;
 use rand::rngs::StdRng;
 use rand::seq::{IteratorRandom, SliceRandom};
 use rand::SeedableRng;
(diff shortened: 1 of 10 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -552,6 +553,22 @@ impl SignatureManager {
...
                 .transact()
                 .await?;
+
+            let status = response.status();
+
+            // Check we returned successfully
+            let FinalExecutionStatus::SuccessValue(payload) = status else {
+                let err = format!("Publish transaction failed, {:?}", response);
+                tracing::error!(err);
+                return Err(near_fetch::Error::RpcReturnedInvalidData(err));
```
volovyks, line 563: We are publishing all signatures in a loop. Do we need to return an Error in case one of the transactions failed?

chain-signatures/contract/src/lib.rs
```diff
@@ -117,6 +117,8 @@ impl MpcContract {
     fn add_sign_result(&mut self, payload: &SignatureRequest, signature: SignatureResponse) {
         if self.pending_requests.contains_key(payload) {
             self.pending_requests.insert(payload, &Some(signature));
+        } else {
+            env::panic_str(&format!("Key not found: {:?}", payload))
```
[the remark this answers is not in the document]
DavidM-D: This needs a test, I'll add it in

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -552,6 +553,22 @@ impl SignatureManager {
...
+            };
+
+            // Check that the return type is correct
```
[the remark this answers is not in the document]
DavidM-D: We don't I'll remove it

chain-signatures/node/src/protocol/signature.rs
DavidM-D: We probably don't, we should probably log the error but not throw anything?

volovyks: That is an option. That is what we are doing now.
Ideally, we will need to add the signature back to the queue to try again in the next round. But only if the error is "recoverable", this one is not.

volovyks: Adding a signature back to the queue can create many edge cases and a need to track one more parameter - signature queue size. I would keep it simple for now and create a separate issue for this if you think it's important.

ChaoticTempest: also, don't need all this custom logic since we're using `near_fetch` here. We can just do
```
if let Err(err) = response.json() {
   ...
}
```
which does both the status checking and serialization step if needed

DavidM-D: Done

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -551,7 +553,25 @@ impl SignatureManager {
                 .max_gas()
                 .retry_exponential(10, 5)
                 .transact()
-                .await?;
+                .await
+            {
+                Ok(response) => response,
+                Err(err) => {
+                    tracing::error!(%receipt_id, error = ?err, "Failed to publish transaction");
+                    break;
```
ChaoticTempest, line 561: We should be doing `continue` instead of `break` since being unable to publish one signature shouldn't be affecting another signature from being published.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -551,7 +553,25 @@ impl SignatureManager {
...
+                }
+            };
+
+            match response.json() {
+                Ok(()) => {
+                    tracing::info!(%receipt_id, bi_r = signature.big_r.affine_point.to_base58(), s = ?signature.s, "published signature sucessfully")
+                }
+                Err(err) => {
+                    tracing::error!(%receipt_id, bi_r = signature.big_r.affine_point.to_base58(), s = ?signature.s, error = ?err, "smart contract threw error");
+                    break;
```
ChaoticTempest, line 571: Same here for this `break`

chain-signatures/contract/src/lib.rs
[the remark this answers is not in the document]
ChaoticTempest: if it's discarded, we need to alert the requestor of the signature that it failed to publish somehow. I guess with yield/resume we can do this by returning an error and pancing so we can do this in future, but maybe for now we just store the signature to be sent again on the next iteration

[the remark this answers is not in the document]
ailisp: For now we can add a record_failure method to contract, and cross contract call to it when it fails. The record_failure can be actually not record anything on chain, but because it is cross contract called, we can modify the indexer part to catch it.

Of course yield/resume would be better if it can be implemented soon, and signature discarded as a short term solution is acceptable.

ailisp (approved):
Looks good to me!

chain-signatures/contract/src/lib.rs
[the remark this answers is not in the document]
DavidM-D: I ended up pushing it to the back of the queue and retrying unless it fails in the contract. It's easy to DoS the contract if you allow incorrect responses to cancel the request.

DavidM-D:
I think this is ready to merge

volovyks (approved):
LGTM

PR #650, opened 2024-06-18
chore: Updated dependencies to node 1.40-rc.1

ChaoticTempest:


volovyks (approved):
LGTM

ppca:
I think according to the error in integration test, we'd need to upload a sandbox tar.gz into the s3 path it wants

ChaoticTempest:
hmm, weird how it inconsistently unavailable. mainnet is now on 1.40. Might as well just go and get the 1.40 binaries instead of this RC one

ppca:
> hmm, weird how it inconsistently unavailable. mainnet is now on 1.40. Might as well just go and get the 1.40 binaries instead of this RC one

we'd need to have sandbox.tar.gz in the 1.40 folder in s3 too if we want it to go thru though...

ChaoticTempest:
yeah, protocol team already added it back. Just tested that it appears in both arm and x86 versions

ChaoticTempest:
Closing this in favor of https://github.com/near/mpc/pull/670

PR #652, opened 2024-06-18
fix: use fixed IPs for dev

ppca:


ChaoticTempest:
how likely do these IPs change?

ChaoticTempest (approved):

kmaus-near (approved):

kmaus-near:
> how likely do these IPs change?

These only change when you destroy and reapply them with Terraform, but to prevent that you can hardcode the IP addresses, so they are always created with the same IP every time.

PR #655, opened 2024-06-24
binary to spin up chain signature local cluster

ailisp:
Spin up a cluster of chain signature nodes for local manual testing, as `with_multichain_nodes` do.  Or spin up only the dependent services but no nodes.


volovyks (dismissed):
LGTM!
Please, add a short instraction in README on how and when we can use this feature.

volovyks (approved):

PR #667, opened 2024-07-03
Test resharing when rpc is congested

ailisp:
Address near/transfer#33. 

Make each node -> RPC with different configurable network problem. (Previously was every node -> same toxiproxy -> lake RPC)

When latency is very high (10s) for a single node but the other node doesn't have latency to RPC, the resharing will fail. Will investigate more on this case.

### DIFF
--- a/integration-tests/chain-signatures/src/containers.rs
+++ b/integration-tests/chain-signatures/src/containers.rs
@@ -30,6 +30,8 @@ pub struct Node<'a> {
     pub cipher_sk: hpke::SecretKey,
     pub sign_pk: near_workspaces::types::PublicKey,
     cfg: MultichainConfig,
+    // near rpc address, after proxy
+    near_rpc: String,
 }
 
 impl<'a> Node<'a> {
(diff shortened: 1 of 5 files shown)


volovyks (approved):
LGTM
Final goal here is to make sure resharing always succeed by waiting for all required conditions which are T or more active nodes with up to date data.

So we need to simulate both cases:
- Node is offline, there are only T-1 nodes and the resharing process is started, all the nodes are waiting for the Node 1 to get back online and finish the resharing process.
- One or more nodes are having issues with getting up to date data from the contract (the nu,ber of active participants is below T). Once T nodes are able to communicate with contract without delays, resharing process should finish successfully.

ppca (approved):
Another thing I've observed while doing contract init (the nodes doing initial key generation), if some node joined network a bit later, e.g. the node started a bit late, then all nodes will be stuck at generating status. Not sure if this is similar to the 10s issue mentioned in description.

ailisp:
> Another thing I've observed while doing contract init (the nodes doing initial key generation), if some node joined network a bit later, e.g. the node started a bit late, then all nodes will be stuck at generating status. Not sure if this is similar to the 10s issue mentioned in description.

Good observation, will try to test and fix this case as well

ailisp:
@volovyks resharing is more important than contract initialization as it's not a one time event. I'll ensure two cases you mentioned are working or fix it

PR #668, opened 2024-07-03
fix: contract init stuck because of secret storage loading

ppca:
secret storage loading value `1`, which is the initial value, returns error instead of None, thus state progression is stuck

### DIFF
--- a/chain-signatures/node/src/storage/secret_storage.rs
+++ b/chain-signatures/node/src/storage/secret_storage.rs
@@ -65,7 +65,14 @@ impl SecretNodeStorage for SecretManagerNodeStorage {
             .load_secret(&self.sk_share_secret_id)
             .await?;
         match raw_data {
-            Some(data) if data.len() > 1 => Ok(Some(serde_json::from_slice(&data)?)),
+            Some(data) if data.len() > 1 => {
+                if let Ok(persistent_node_data) = serde_json::from_slice(&data) {
+                    Ok(Some(persistent_node_data))
+                } else {
+                    tracing::info!("failed to convert stored data to key share, presuming it is missing");
+                    Ok(None)
+                }
+            },
             _ => {
                 tracing::info!("failed to load existing key share, presuming it is missing");
                 Ok(None)


chain-signatures/node/src/storage/secret_storage.rs
```diff
@@ -65,7 +65,14 @@ impl SecretNodeStorage for SecretManagerNodeStorage {
...
             .await?;
         match raw_data {
-            Some(data) if data.len() > 1 => Ok(Some(serde_json::from_slice(&data)?)),
+            Some(data) if data.len() > 1 => {
+                if let Ok(persistent_node_data) = serde_json::from_slice(&data) {
+                    Ok(Some(persistent_node_data))
+                } else {
+                    tracing::info!("failed to convert stored data to key share, presuming it is missing");
+                    Ok(None)
+                }
```
ChaoticTempest, line 74: we should log the error just in case so we can debug what the exact cause was. We shouldn't include the actual `data` portion itself but should include things like the length of it just in case we made some incompatible version serialization changes to persistent node data

chain-signatures/node/src/storage/secret_storage.rs
```diff
@@ -65,7 +65,14 @@ impl SecretNodeStorage for SecretManagerNodeStorage {
             .load_secret(&self.sk_share_secret_id)
             .await?;
         match raw_data {
-            Some(data) if data.len() > 1 => Ok(Some(serde_json::from_slice(&data)?)),
+            Some(data) if data.len() > 1 => {
+                if let Ok(persistent_node_data) = serde_json::from_slice(&data) {
+                    Ok(Some(persistent_node_data))
+                } else {
+                    tracing::info!("failed to convert stored data to key share, presuming it is missing");
```
ailisp, line 72:
```suggestion
                    tracing::warn!("failed to convert stored data to key share, presuming it is missing");
```

chain-signatures/node/src/storage/secret_storage.rs
ailisp: I agree that if we want to debug in case of incompatible change, we need to log. But we still need to log or store actual data somewhere, because old storage would get overwritten

ailisp: Good point on the case of debugging. I suggest to make the log level to warn. Current log information should be sufficient to debug (timestamp, node config's sk_share_secret_id), make it to warn maybe easier to track unexpected case in monitor.

ailisp (approved):

volovyks (approved):

PR #671, opened 2024-07-08
fix: RUSTSEC-2024-0344

ChaoticTempest:
This fixes the vulnerability for RUSTSEC-2024-0344. But ignores two other ones due to it only affecting the CLI and workspaces testing, which isn't vital to deal with for now.


chain-signatures/Cargo.toml
```diff
@@ -5,3 +5,9 @@ members = [
     "node",
 ]
 resolver = "2"
+
+[patch.crates-io]
+# TODO: trigger Cargo.lock update for x25519-dalek once they release.
+# This fixes https://rustsec.org/advisories/RUSTSEC-2024-0344 by pointing to a commit that includes the fix.
+# This fix has yet to be propagated to crates.io so we will patch it instead.
+x25519-dalek = { git = "https://github.com/dalek-cryptography/curve25519-dalek", rev = "5b7082bbc8e0b2106ab0d956064f61fa0f393cdc" }
```
ailisp, line 13: Nice fix! Should we also patch it for fastauth

ChaoticTempest: Nope, fastauth should just be left as-is for now so that we don't break anything

ChaoticTempest:
This PR now also ignores RUSTSEC-2024-0347 due to it only affecting integration tests

ppca (approved):

ppca (approved):

PR #676, opened 2024-07-10
feat: Added nightly test that runs at 3:00 every day

ChaoticTempest:
This adds a new nightly GH actions test that schedules a run at 3:00 am every single day. It utilizes an 8 node setup that will produce 100 signatures. The amount of signatures produced is low for now to test it out the nightly run.

GH actions might not be able to handle the 8 node setup, so will backpedal to 5 or something lower if needed later.


integration-tests/chain-signatures/tests/cases/nightly.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    with_multichain_nodes(config, |ctx| {
+        Box::pin(async move {
+            let state_0 = wait_for::running_mpc(&ctx, Some(0)).await?;
+            assert_eq!(state_0.participants.len(), NODES);
+            wait_for::has_at_least_mine_triples(&ctx, triple_cfg.min_triples).await?;
+
+            for _ in 0..SIGNATURE_AMOUNT {
+                wait_for::has_at_least_mine_presignatures(&ctx, 3).await?;
+                actions::single_signature_production(&ctx, &state_0).await?;
```
ailisp, line 50: Maybe not await but collect no. of success and no. of failures (& reason)

ailisp (dismissed):
looks good to me!

integration-tests/chain-signatures/tests/cases/nightly.rs
ChaoticTempest: True, I'll add error logs to this

volovyks (commented):
Let's establish what we want to achieve with this test. It is still synthetic because we await triples before each signature request. Even if we remove those checks, we will still execute them one by one.

In real life, requests come in batches without knowledge about our system state. That is why I advocate using existing load tests. We can create any scenarios by modifying a simple config file and observing statistics in the end (if we save it somewhere or log it at the end of the test).

ChaoticTempest:
Will add something more akin to load tests to these nightlies in the future. Adding this as purely setup for now

ppca (approved):

PR #681, opened 2024-07-12
feat: stuck protocol monitor

ChaoticTempest:
This will allow to keep tabs on whether our triple protocols are stuck or not while testing. This will be vital for monitoring if there is something going wrong or not. By default, we will report that the protocol is stuck if the amount of triples remain the same `STUCK_TIMEOUT_THRESHOLD` seconds. There's a `STUCK_COUNT_REPORT_INTERVAL` that will report the stuckness of the system every so often just to not spam the logs.

Ideally should make it into grafana metrics later

### DIFF
--- a/chain-signatures/node/src/protocol/monitor.rs
+++ b/chain-signatures/node/src/protocol/monitor.rs
@@ -1,0 +1,74 @@
+use std::collections::HashSet;
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+use super::triple::{TripleId, TripleManager};
+
+/// Amount of iterations before we can say that the protocol is stuck.
+const ADVANCEMENT_THRESHOLD: usize = 20;
+
+pub struct StuckMonitor {
+    triple_manager: Arc<RwLock<TripleManager>>,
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/protocol/monitor.rs
```diff
@@ -0,0 +1,77 @@
...
+            return;
+        }
+
+        if self.last_changed_timestamp.elapsed() >= STUCK_TIMEOUT_THRESHOLD
+            && self.stuck_interval_timestamp.elapsed() >= STUCK_COUNT_REPORT_INTERVAL
+        {
+            self.stuck_interval_timestamp = Instant::now();
+
+            tracing::warn!(
+                // ?latest_triples,
```
volovyks, line 62: Do we need this?

volovyks (approved):

chain-signatures/node/src/protocol/monitor.rs
ChaoticTempest: nope, removed it

volovyks (approved):

PR #690, opened 2024-07-17
contract: return error rather than panic whenever possible

ppca:
all fn will return Result<_, MpcContractError>.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -122,7 +122,6 @@ pub enum ThresholdError {
     TooHigh,
 }
 
-
 #[derive(Debug, thiserror::Error)]
 pub enum VoteError {
     #[error("Voting account is not not in the participant set.")]
@@ -250,7 +249,7 @@ impl VersionedMpcContract {
         assert!(
             deposit.as_yoctonear() >= required_deposit,
             "Insufficient deposit attached. Attached deposit: {}, required deposit: {}",
-            deposit, 
+            deposit,
             required_deposit
         );
         // Make sure sign call will not run out of gas doing recursive calls because the payload will never be removed
@@ -705,7 +704,10 @@ impl VersionedMpcContract {
 impl VersionedMpcContract {
     #[handle_result]
     #[init]
-    pub fn init(threshold: usize, candidates: BTreeMap<AccountId, CandidateInfo>) -> Result<Self, MpcContractError> {
+    pub fn init(
+        threshold: usize,
+        candidates: BTreeMap<AccountId, CandidateInfo>,
+    ) -> Result<Self, MpcContractError> {
         log!(
             "init: signer={}, threshold={}, candidates={}",
             env::signer_account_id(),
@@ -714,7 +716,7 @@ impl VersionedMpcContract {
         );
 
         if threshold > candidates.len() {
-            return Err(MpcContractError::ThresholdError(ThresholdError::TooHigh))
+            return Err(MpcContractError::ThresholdError(ThresholdError::TooHigh));
         }
 
         Ok(Self::V0(MpcContract::init(threshold, candidates)))
@@ -740,7 +742,7 @@ impl VersionedMpcContract {
         );
 
         if threshold > participants.len() {
-            return Err(MpcContractError::ThresholdError(ThresholdError::TooHigh))
+            return Err(MpcContractError::ThresholdError(ThresholdError::TooHigh));
         }
 
         Ok(Self::V0(MpcContract {
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -1,3 +1,4 @@
+pub mod errors;
 pub mod primitives;
 
 use crypto_shared::{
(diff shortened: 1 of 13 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -320,39 +387,50 @@ impl VersionedMpcContract {
...
         let epsilon = derive_epsilon(&predecessor, &path);
         let derived_public_key =
-            derive_key(near_public_key_to_affine_point(self.public_key()), epsilon);
+            derive_key(near_public_key_to_affine_point(self.public_key()?), epsilon);
         let encoded_point = derived_public_key.to_encoded_point(false);
         let slice: &[u8] = &encoded_point.as_bytes()[1..65];
         let mut data: Vec<u8> = vec![near_sdk::CurveType::SECP256K1 as u8];
         data.extend(slice.to_vec());
-        PublicKey::try_from(data).unwrap()
+        Ok(PublicKey::try_from(data).unwrap())
```
ChaoticTempest, line 433: we should map this unwrap to an error on the off chance that the derive function doesn't work

chain-signatures/contract/tests/tests.rs
```diff
@@ -251,7 +251,7 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
     let err = sign_and_validate(&request, None, &contract)
         .await
         .expect_err("should have failed with timeout");
-    assert!(err.to_string().contains("Signature has timed out"));
+    assert!(err.to_string().contains("Signature request has timed out"));
```
ChaoticTempest, line 254: we should be able to do `contains(Error.to_string())` since they implement `Display`

frol:
@ppca @ChaoticTempest @volovyks Sorry for jumping in, but I'd like to invite you to participate in the [Race of Sloths](https://race-of-sloths.com) - a fun and gamified open-source contributions program. Consider mentioning @race-of-sloths user in your github comment or PR description to join!

See how the flow works here: https://github.com/near/nearcore/pull/11778

ailisp (commented):
Changes to the contract looks good overall, 

but (If near-sdk-rs macro doesn't turn Err into a failed transaction), integration tests and node will need to adapt to the return error version of the contract. 

For example:

```rs
    futures::future::join_all(vote_futures)
        .await
        .iter()
        .for_each(|result| {
            assert!(result.as_ref().unwrap().failures().is_empty());
        });
```

assert there is no failure in `vote_join` calls, but now vote_join will always be successful transactions (returning Ok or Err), so this will always pass. Same for vote_leave.

And same for mpc node calling the contract, e.g.:
```rs

                        ctx.rpc_client()
                            .call(ctx.signer(), ctx.mpc_contract_id(), "join")
                            .args_json(json!({
                                "url": ctx.my_address(),
                                "cipher_pk": ctx.cfg().network_cfg.cipher_pk.to_bytes(),
                                "sign_pk": ctx.cfg().network_cfg.sign_sk.public_key(),
                            }))
                            .max_gas()
                            .retry_exponential(10, 3)
                            .transact()
                            .await
                            .map_err(|err| {
                                tracing::error!(?err, "failed to join the participant set");
                                ConsensusError::CannotJoin(format!("{err:?}"))
                            })?;
```
will not enter map_err since it will be `Ok(ExecutionFinalResult)`

ailisp (approved):

chain-signatures/contract/src/lib.rs
```diff
@@ -94,6 +95,62 @@ pub struct YieldIndex {
...
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum SignatureError {
+    #[error("Signature request has timed out.")]
+    Timeout,
+    #[error("This sign request was removed from pending requests: timed out or completed.")]
+    RequestNotInPending,
+    #[error("Signature could not be verified.")]
+    SignatureNotVerified,
```
volovyks, line 105: Can we please separate errors by function or scope where they can appear?
The end goal here is to give users a finite number of errors that can happen for each call they are making.

In this case, we have `Timeout`, which the user can get from `sign()` and `SignatureNotVerified`, which can only happen in `respond`.

chain-signatures/contract/src/lib.rs
```diff
@@ -130,11 +187,14 @@ impl MpcContract {
         }
     }
 
-    fn remove_request(&mut self, request: SignatureRequest) {
+    fn remove_request(&mut self, request: SignatureRequest) -> Result<(), MpcContractError> {
```
volovyks, line 190: Is it possible to add a similar return Error type for `sign()`? This is the main function and it is a shame that it does not have strongly typed errors.

volovyks (commented):
Looks cool!

chain-signatures/contract/src/lib.rs
```diff
@@ -202,12 +270,12 @@ impl VersionedMpcContract {
...
-                "Attached deposit is {}, required deposit is {}",
-                deposit, required_deposit
-            ));
-        }
+        assert!(
+            deposit.as_yoctonear() >= required_deposit,
+            "Insufficient deposit attached. Attached deposit: {}, required deposit: {}",
+            deposit,
+            required_deposit
+        );
```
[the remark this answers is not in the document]
ppca: no I tried, it doesn't work. It will return () instead of the actual yield/resume result. Let me ask Saketh if it's possible.

chain-signatures/contract/src/lib.rs
ppca: sounds good. Will do.

chain-signatures/contract/src/lib.rs
ppca: no I tried, it doesn't work. It will return () instead of the actual yield/resume result. Let me ask Saketh if it's possible.

ppca:
using the trick of mimicking cross contract call to return a promise from the yield/resume

ChaoticTempest (approved):
not a huge fan of adding another promise on top just to have the return type, but we can probably fix this later without breaking anything since the return promise isn't directly used, just dropped to invoke the promise call

PR #691, opened 2024-07-18
contract: add my_derived_public_key()

ppca:


chain-signatures/contract/src/lib.rs
```diff
@@ -373,6 +374,19 @@ impl VersionedMpcContract {
         }
     }
 
+    /// This is the derived key given path
+    pub fn my_derived_public_key(&self, path: String) -> PublicKey {
```
ChaoticTempest, line 378: Don't think `my` adds any more info here, and we should just keep consistent with verbs in contract interfaces. How about we just rename to `derive_public_key`

ppca: changed it to derived_public_key() to be consistent with public_key()

chain-signatures/contract/src/lib.rs
```diff
@@ -373,6 +374,19 @@ impl VersionedMpcContract {
         }
     }
 
+    /// This is the derived public key of the caller given path
+    pub fn derived_public_key(&self, path: String) -> PublicKey {
+        let predecessor = env::predecessor_account_id();
```
ChaoticTempest, line 379: I think we should add in an option where people can specify the predecessor account id since the one calling this function might or might not be the right one. Like in this case:
```
user.near calls nft.near
nft.near calls sign
predecessor_id = nft.near

user.near calls derived_public_key
predecessor_id = user.near
```
```suggestion
    pub fn derived_public_key(&self, path: String, predecessor: Option<AccountId>) -> PublicKey {
        let predecessor = predecessor.unwrap_or_else(env::predecessor_account_id);
```

ChaoticTempest (approved):

chain-signatures/contract/src/lib.rs
ppca: sweet. Will do.

ChaoticTempest (approved):

PR #695, opened 2024-07-18
mainnet partner instruction: modify script to generate near account keys

ppca:
This is for creating the mainnet near account partners would use for their nodes.


ChaoticTempest (approved):
Is there a way where node providers can run this script while also setting these values in their secret storage on GCP? That way they don't have to manually copy and paste these values

ppca:
> Is there a way where node providers can run this script while also setting these values in their secret storage on GCP? That way they don't have to manually copy and paste these values

I think Kody has some script that might be able to do it. I'll ask

PR #701, opened 2024-07-19
Migrate to sha3 for the KDF

DavidM-D:
From the audit, not completely convinced that it's necessary, but better to be safe than sorry.


volovyks (dismissed):
LGTM

volovyks (approved):

PR #703, opened 2024-07-19
Fail on bad scalar conversions

### LINKED ISSUE #702: KS–PCS–F–7 Scalars Initiated with Unsafe Function

DavidM-D:
Previously we'd allow bad scalars to be converted too and from bytes. This should be impossible since all conversions are done internally, but it's good to check.

Closes #702

### DIFF
--- a/chain-signatures/contract/tests/tests.rs
+++ b/chain-signatures/contract/tests/tests.rs
@@ -1,12 +1,13 @@
 use crypto_shared::kdf::{check_ec_signature, derive_secret_key};
 use crypto_shared::{
-    derive_epsilon, derive_key, SerializableAffinePoint, SerializableScalar, SignatureResponse,
+    derive_epsilon, derive_key, ScalarExt as _, SerializableAffinePoint, SerializableScalar,
+    SignatureResponse,
 };
 use ecdsa::signature::Verifier;
 use k256::elliptic_curve::ops::Reduce;
 use k256::elliptic_curve::point::DecompressPoint;
 use k256::elliptic_curve::sec1::ToEncodedPoint;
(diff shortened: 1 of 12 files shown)


DavidM-D:
Is it OK that this panics, or should I get it to return a Result?

volovyks:
We were using the unsafe code to process data provided by the user. As a result, there were no guarantees about the integrity of the data usually provided for an unsafe block of code.

Now we are changing that to a `panic()` but it is still user-provided data that can crash the program. I would add error handling here, but I see your point, it requires a lot of refactoring.
cc @ChaoticTempest take a look, please.

DavidM-D:
I'll make it an option, it's actually showing some interesting parts of our codebase. Like what happens if, when deriving epsilon, the output of the hash function doesn't fall into the field?

chain-signatures/contract/tests/tests.rs
```diff
@@ -135,8 +136,8 @@ async fn create_response(
 
     let s = signature.s();
     let (r_bytes, _s_bytes) = signature.split_bytes();
-
-    let respond_req = SignatureRequest::new(payload_hash, predecessor_id, path);
+    let payload_hash_s = Scalar::from_bytes(payload_hash).unwrap();
+    let respond_req = SignatureRequest::new(payload_hash_s, predecessor_id, path);
```
ChaoticTempest, line 140: you don't need to do the conversion. There's already `scalar_hash` above

ChaoticTempest (approved):
It's fine for now to panic until we get error handling into the contract. At least with this while deriving epsilon, it panics at the entrypoint in the contract before making it to the MPC network. It's kinda buried where this guard is happening but we will have to surface it in a later error handing PR

PR #706, opened 2024-07-21
Contract refactoring, API, README, etc

volovyks:
- fixed contract tests
- renamed some errors and error messages (opinionated, open for a discussion about some of them)
- moved user-facing functions into one impl block
- added README with an API and a list of envs (mostly existing descriptions moved to a text file and formatted for better onboarding).

I've also tried to narrow down return error types in user-facing functions so they know what they can get exactly.
```rust
/// from
pub fn sign(&mut self, request: SignRequest) -> Result<near_sdk::Promise, MpcContractError>
/// to
pub fn sign(&mut self, request: SignRequest) -> Result<near_sdk::Promise, SignError>
```
But to do that we will need to get rid of the MpcContractError enum since it's implementation of 
```rust
impl near_sdk::FunctionError for MpcContractError {
    fn panic(&self) -> ! {
        crate::env::panic_str(&self.to_string())
    }
}
```
conflicts with newly added impl for SignError, PublicKeyError, etc. 
Let me know if you guys think this is an important change, or we can stick to MpcError everyvere.

@ppca based on your branch, since it has a lot of changes


chain-signatures/contract/src/errors.rs
```diff
@@ -1,36 +1,40 @@
...
+    UnsupportedKeyVersion,
+    #[error("Attached deposit is lower than required. Attached: {0}, Required: {1}.")]
+    InsufficientDeposit(u128, u128),
+    #[error("Provided gas is lower than required. Provided: {0}, required {1}.")]
+    InsufficientGas(Gas, Gas),
+    #[error("Too many pending requests. Please try again later.")]
     RequestLimitExceeded,
-    #[error("This sign request was removed from pending requests: timed out or completed.")]
-    RequestNotInPending,
+    #[error("This sign request had timed out, was completed, or never existed.")]
```
DavidM-D, line 21:
```suggestion
    #[error("This sign request has timed out, was completed, or never existed.")]
```

chain-signatures/contract/src/errors.rs
```diff
@@ -1,36 +1,40 @@
...
 
 #[derive(Debug, thiserror::Error)]
 pub enum RespondError {
-    #[error("This sign request was removed from pending requests: timed out or completed.")]
-    RequestNotInPending,
-    #[error("Signature could not be verified.")]
-    SignatureNotVerified,
-    #[error("Protocol state is not running.")]
-    ProtocolStateNotRunning,
+    #[error("This sign request had timed out, was completed, or never existed.")]
```
DavidM-D, line 27:
```suggestion
    #[error("This sign request has timed out, was completed, or never existed.")]
```

chain-signatures/contract/src/errors.rs
```diff
@@ -1,36 +1,40 @@
...
+    RequestNotFound,
+    #[error("The provided signature is invalid.")]
+    InvalidSignature,
+    #[error("The protocol is not in the Running state.")]
```
DavidM-D, line 31:
```suggestion
    #[error("The protocol is not Running.")]
```

chain-signatures/contract/src/errors.rs
```diff
@@ -1,36 +1,40 @@
...
+    ProtocolNotInRunningState,
 }
 
 #[derive(Debug, thiserror::Error)]
 pub enum JoinError {
-    #[error("Protocol state is not running")]
+    #[error("The protocol is not in the Running state.")]
```
DavidM-D, line 37:
```suggestion
    #[error("The protocol is not Running.")]
```

chain-signatures/contract/src/lib.rs
```diff
@@ -712,6 +611,80 @@ impl VersionedMpcContract {
         }
     }
 
+    // contract version
+    pub fn version(&self) -> String {
+        env!("CARGO_PKG_VERSION").to_string()
+    }
+
+    pub fn sign_helper(&mut self, request: SignatureRequest) {
```
gagdiez, line 619: @volovyks should this be marked as `[#private]`?

otherwise, people could just call it jumping over `sign` (except if there is something about using the `DATA_ID_REGISTER` that I am not aware of)

chain-signatures/contract/src/lib.rs
```diff
@@ -179,23 +144,28 @@ impl VersionedMpcContract {
...
         let deposit = env::attached_deposit();
         let required_deposit = self.signature_deposit();
         if deposit.as_yoctonear() < required_deposit {
-            return Err(MpcContractError::SignError(SignError::DepositInsufficient(
-                "{required_deposit}".to_string(),
+            return Err(MpcContractError::SignError(SignError::InsufficientDeposit(
+                deposit.as_yoctonear(),
+                required_deposit,
             )));
         }
```
gagdiez, line 163: could we return the money if they exceed the payment?

```
if required > deposit {
  Promise::new(env::predecessor_account_id()).transfer(deposit - required_deposit);} 
}
```

gagdiez:
@volovyks the current cost is 1 yoctoNEAR when there are less than 4 requests, and then it ramps to `50milliNEAR * (1...4)`, which ranges from `0.05N` to `0.20N`... this might be a bit expensive.

Did we intended to use `yoctoNear` instead of milliNEAR?
  
```rust
match pending_requests {
    0..=CHEAP_REQUESTS => 1,
    _ => {
        (pending_requests - CHEAP_REQUESTS) as u128
            * NearToken::from_millinear(50).as_yoctonear()
    }
}
```

volovyks:
Audit is fixed in 

> @volovyks the current cost is 1 yoctoNEAR when there are less than 4 requests, and then it ramps to `50milliNEAR * (1...4)`, which ranges from `0.05N` to `0.20N`... this might be a bit expensive.
> 
> Did we intended to use `yoctoNear` instead of milliNEAR?
> 
> ```rust
> match pending_requests {
>     0..=CHEAP_REQUESTS => 1,
>     _ => {
>         (pending_requests - CHEAP_REQUESTS) as u128
>             * NearToken::from_millinear(50).as_yoctonear()
>     }
> }
> ```

We use yocto everywhere.
It may be expensive, but the intention is to prevent system overload. Attacks must be expensive in order to prevent them.

ppca (approved):

PR #709, opened 2024-07-22
reduce downtime: use newest instead of oldest triple and presig

ppca:
Now when we need to generate presig, we take oldest triples first; when generating signature, we take oldest presig first. However when some nodes restart with no presig and other nodes keep popping old presig, all signature requests will time out until all the old presigs are exhausted. So it's better to start with newer presigs, so as more presig get generated, signature requests will be served in time.

### DIFF
--- a/chain-signatures/Cargo.lock
+++ b/chain-signatures/Cargo.lock
@@ -1753,6 +1753,7 @@ dependencies = [
  "near-sdk",
  "serde",
  "serde_json",
+ "sha3",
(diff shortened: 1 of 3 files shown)


volovyks:
Can this cause other issues? Will we waste more triple because they were "outdated"? Can it cause a situation where we have 1000 triples, but most of them will be discarded since they are old? And such a situation can happen when we need them the most. When our system is under load.

Issue #680, opened 2024-07-11
Include node heights metric in the liveliness of a node


ChaoticTempest:
We should utilize the node's block height as a sign of the node being alive or not. If the node is severely behind in block height, we should say that it is offline, since it is not useable as a participant currently to serve signature requests.

An honest node can self report themselves by returning `NotRunning` when requested on the `/state` endpoint. This is how we can detect our own node is behind:
```py
error_margin = 25
is_behind = nodes[self].height < (avg_height(nodes[0..n] - nodes[self]) - error_margin)
```

It's a bit harder for non-honest nodes, where our own node has to mark other nodes as offline due to their block heights. This will require a little more thought on how to resolve and more research into how NEAR nodes deal with nodes being behind. One naive solution is to check the nodes height, and run an outlier detection algorithm on top of the series of heights. Something like this, where we get the bottom 10% and top 90% of the data.
```py
heights = heights[(heights < np.quantile(heights, 0.1)) & (heights > np.quantile(heights, 0.9))]
```
And then mark the corresponding nodes as offline from the our own node's viewpoint. This definitely requires more testing due to the nature of how all nodes will be doing this offline node assessment, where a general consensus has to be met.

Also, if the network is congested, all nodes should roughly be behind ideally so this should not be an issue in that case.

ChaoticTempest:
Also note that even if a node is offline, it can still be a participant for triple and presignature generation. It just cannot be a part of signature production since signature production requires the indexer to be roughly in line with everyone else to process a signature request.

volovyks:
@ChaoticTempest does it mean that when N=8 and T=5 we have 6 online nodes, it is better to add everybody as active participants in the triple generation protocol?

ChaoticTempest:
not sure what you mean by that example. Currently, how we tell a node is online or not is through calling into `/state` on that particular node which tells us that the node is online. But that has issues when that node's is unable to process signature requests when their indexer is offline. So that means that in our liveness of nodes, it would still get counted as online because it was still able to serve `/state`.

What I'm proposing is to include the block height value in our `/state` calls to tell whether or not a node is running, or nodes themselves can mark themselves as offline by not responding in `/state`.

PR #710, opened 2024-07-23
feat: add latest_block_height and is_stable to /state; only stable nodes can partake in signature generation

ChaoticTempest:
This adds in the `latest_block_height` and `is_stable` values to `StateView` so a node can keep track whether a connection is stable or not. A node will only consider another node to be able to generate a signature if and only if it is stable, such as when it is caught up with indexed blocks. This only affects signature generation, so things like triples, presginatures, resharing are still fine even if a node is not stable.

This fixes https://github.com/near/mpc/issues/680 partially but only handles the honest node portion.

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -97,10 +97,17 @@ impl Indexer {
         }
     }
 
+    /// Get the latest block height from the chain.
     pub async fn latest_block_height(&self) -> BlockHeight {
         self.latest_block_height.read().await.block_height
     }
 
+    /// Check whether the indexer is on track with the latest block height from the chain.
+    pub async fn is_on_track(&self) -> bool {
+        self.last_updated_timestamp.read().await.elapsed() <= Self::BEHIND_THRESHOLD
+    }
+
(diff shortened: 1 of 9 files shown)


chain-signatures/node/src/indexer.rs
```diff
@@ -78,13 +81,59 @@ pub struct ContractSignRequest {
...
+            last_updated_timestamp: Arc::new(RwLock::new(Instant::now())),
+        }
+    }
+
+    /// Get the latest block height from the chain.
+    pub async fn latest_block_height(&self) -> BlockHeight {
+        self.latest_block_height.read().await.block_height
+    }
+
+    /// Check whether the indexer is on track with the latest block height from the chain.
```
volovyks, line 105: Do we need both is_on_track and is behind?

volovyks (approved):
LGTM.
I was expecting that nodes would fetch other nodes' states and decide if they are active by themselves, but that approach has its downsides too.

chain-signatures/node/src/indexer.rs
ChaoticTempest: nope, but nice to have

PR #714, opened 2024-07-24
feat: contract is now upgradeable through participants voting for updates

ChaoticTempest:
The contract can now be upgraded or config updated through calling into:
- `propose_update: (config | wasm) -> UpdateId`
- `vote_update: (UpdateId) -> bool`

Only participants from the contract are able to partake. Added config here just to test it can work there too.

### DIFF
--- a/chain-signatures/contract/src/config.rs
+++ b/chain-signatures/contract/src/config.rs
@@ -1,0 +1,30 @@
+use borsh::{self, BorshDeserialize, BorshSerialize};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+pub struct Config {
+    /// Timeout for triple generation in milliseconds.
+    pub triple_timeout: u64,
+    /// Timeout for presignature generation in milliseconds.
+    pub presignature_timeout: u64,
+    /// Timeout for signature generation in milliseconds.
+    pub signature_timeout: u64,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            triple_timeout: min_to_ms(20),
+            presignature_timeout: secs_to_ms(30),
+            signature_timeout: secs_to_ms(30),
+        }
+    }
+}
+
+pub const fn secs_to_ms(secs: u64) -> u64 {
+    secs * 1000
+}
+
+pub const fn min_to_ms(min: u64) -> u64 {
+    min * 60 * 1000
+}
--- a/chain-signatures/contract/src/errors.rs
+++ b/chain-signatures/contract/src/errors.rs
@@ -66,8 +66,18 @@ pub enum VoteError {
     EpochMismatch,
     #[error("Number of participants cannot go below threshold.")]
     ParticipantsBelowThreshold,
+    #[error("Update not found.")]
+    UpdateNotFound,
     #[error("Unexpected protocol state: {0}")]
     UnexpectedProtocolState(String),
+    #[error("Unexpected: {0}")]
+    Unexpected(String),
+}
+
+impl near_sdk::FunctionError for VoteError {
+    fn panic(&self) -> ! {
+        crate::env::panic_str(&self.to_string())
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -77,7 +87,7 @@ pub enum MpcContractError {
     #[error("respond fn error: {0}")]
     RespondError(RespondError),
     #[error("vote_* fn error: {0}")]
-    VoteError(VoteError),
+    VoteError(#[from] VoteError),
     #[error("init fn error: {0}")]
     InitError(InitError),
     #[error("join fn error: {0}")]
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -1,29 +1,32 @@
 pub mod errors;
 pub mod primitives;
+pub mod state;
 
 use crypto_shared::{
     derive_epsilon, derive_key, kdf::check_ec_signature, near_public_key_to_affine_point,
     types::SignatureResponse, ScalarExt as _,
 };
+use k256::elliptic_curve::sec1::ToEncodedPoint;
 use k256::Scalar;
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::collections::LookupMap;
-use near_sdk::serde::{Deserialize, Serialize};
-
 use near_sdk::{
     env, log, near_bindgen, AccountId, CryptoHash, Gas, GasWeight, NearToken, PromiseError,
     PublicKey,
 };
-
-use errors::{
-    InitError, JoinError, MpcContractError, PublicKeyError, RespondError, SignError, VoteError,
-};
-use k256::elliptic_curve::sec1::ToEncodedPoint;
 use primitives::{
     CandidateInfo, Candidates, Participants, PkVotes, SignRequest, SignaturePromiseError,
     SignatureRequest, SignatureResult, StorageKey, Votes, YieldIndex,
 };
 use std::collections::{BTreeMap, HashSet};
+
+use crate::errors::{
+    InitError, JoinError, MpcContractError, PublicKeyError, RespondError, SignError, VoteError,
+};
+
+pub use state::{
+    InitializingContractState, ProtocolContractState, ResharingContractState, RunningContractState,
+};
 
 const GAS_FOR_SIGN_CALL: Gas = Gas::from_tgas(250);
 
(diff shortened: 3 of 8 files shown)


chain-signatures/contract/src/config.rs
```diff
@@ -0,0 +1,30 @@
+use borsh::{self, BorshDeserialize, BorshSerialize};
+use serde::{Deserialize, Serialize};
+
+#[derive(Clone, Debug, Serialize, Deserialize, BorshSerialize, BorshDeserialize, PartialEq, Eq)]
+pub struct Config {
```
volovyks, line 5: We can add a config that @DavidM-D was talking about — parameters for a function that calculates the required deposit. Conversely, we can hardcode all the parameters and reduce complexity. Update the contract each time we need to change something. It is a bit more risky though.

chain-signatures/contract/src/lib.rs
```diff
@@ -533,15 +501,74 @@ impl VersionedMpcContract {
...
+
+    /// Propose an update to the contract. [`Update`] are all the possible updates that can be proposed.
+    ///
+    /// returns Some(id) if the proposal was successful, None otherwise
+    #[handle_result]
+    pub fn propose_update(
+        &mut self,
+        code: Option<Vec<u8>>,
+        config: Option<Config>,
+    ) -> Result<UpdateId, VoteError> {
```
volovyks, line 522: I was asking about this pattern in one of my previous PRs. Does it make sense to return more specific errors on all, or all user-facing functions instead of the generic MPC error?

chain-signatures/contract/src/errors.rs
```diff
@@ -77,7 +87,7 @@ pub enum MpcContractError {
     #[error("respond fn error: {0}")]
     RespondError(RespondError),
     #[error("vote_* fn error: {0}")]
-    VoteError(VoteError),
+    VoteError(#[from] VoteError),
```
volovyks, line 90: Does this help with conflicting `impl near_sdk::FunctionError for VoteError`?

chain-signatures/contract/src/lib.rs
```diff
@@ -533,15 +501,74 @@ impl VersionedMpcContract {
...
+        };
+
+        // Not enough votes, wait for more.
+        if votes.len() < threshold {
+            return Ok(false);
+        }
+
+        let Some(_promise) =
+            self.proposed_updates()
+                .do_update(&id, "update_config", UPDATE_CONFIG_GAS)
```
volovyks, line 565: Do we need to pass the name of the function here? It is used only in one place.

chain-signatures/contract/tests/tests.rs
```diff
@@ -371,3 +396,144 @@ async fn test_contract_initialization() -> anyhow::Result<()> {
...
+    test_propose_update_config(&contract, &accounts).await;
+    test_propose_update_contract(&contract, &accounts).await;
+}
+
+async fn test_propose_update_config(contract: &Contract, accounts: &[Account]) {
+    // contract should not be able to propose updates unless its apart of the participant/voter set.
+    let execution = contract
+        .call("propose_update")
+        .args_json(serde_json::json!({
+            "contract": vec![1, 2, 3],
```
volovyks, line 414: The options are `code` or `config`.

chain-signatures/contract/tests/tests.rs
```diff
@@ -371,3 +396,144 @@ async fn test_contract_initialization() -> anyhow::Result<()> {
...
+        }))
+        .transact()
+        .await
+        .unwrap();
+    dbg!(&execution);
+    assert!(execution.is_failure());
```
volovyks, line 420: Not blocking this PR, but since we have a unified way to check if acc can vote, it would be great to check for an error type explicitly.

chain-signatures/contract/tests/tests.rs
```diff
@@ -371,3 +396,144 @@ async fn test_contract_initialization() -> anyhow::Result<()> {
...
+
+    // have each participant propose a new update:
+    let new_config = Config {
+        triple_timeout: min_to_ms(10),
+        ..Default::default()
+    };
+    let mut proposals = Vec::with_capacity(accounts.len());
+    for account in accounts {
```
volovyks, line 428: I assume all of the participants will get different IDs even if the proposal is the same.
That is fine with me, just trying to understand the logic here.

chain-signatures/contract/tests/tests.rs
```diff
@@ -371,3 +396,144 @@ async fn test_contract_initialization() -> anyhow::Result<()> {
...
+    assert_ne!(config, old_config);
+    assert_eq!(config, new_config);
+}
+
+async fn test_propose_update_contract(contract: &Contract, accounts: &[Account]) {
+    let state: mpc_contract::ProtocolContractState =
+        contract.view("state").await.unwrap().json().unwrap();
+
+    // Let's propose a contract update instead now.
+    let new_wasm = std::fs::read(CONTRACT_FILE_PATH).unwrap();
```
volovyks, line 486: Technically, it is the same contract.
Not blocking, but we will need to have a test, that the contract was updated. More than that, it would be great to have a test, that this new contract is also upgradable :)

chain-signatures/contract/src/lib.rs
```diff
@@ -347,6 +324,7 @@ impl VersionedMpcContract {
             env::signer_account_id(),
             candidate_account_id
         );
+        let voter = self.voter()?;
```
volovyks, line 327: Looks great!

chain-signatures/contract/src/lib.rs
```diff
@@ -533,15 +501,74 @@ impl VersionedMpcContract {
...
                 } else {
                     Err(MpcContractError::VoteError(
-                        VoteError::UnexpectedProtocolState("resharing".to_string()),
+                        VoteError::UnexpectedProtocolState("Running".to_string()),
                     ))
                 }
             }
             _ => Err(MpcContractError::VoteError(
-                VoteError::UnexpectedProtocolState("resharing".to_string()),
+                VoteError::UnexpectedProtocolState("Running".to_string()),
```
volovyks, line 509: I agree that this one is not `resharing`, but it is not `running` either.

chain-signatures/contract/src/lib.rs
volovyks, line 565: This looks a bit odd. You are passing "update_config", but this function will both update the config and do the update.

volovyks (commented):
Looks awesome! And delivered in 1 day, cool.

chain-signatures/contract/src/config.rs
```diff
@@ -0,0 +1,30 @@
...
+    /// Timeout for presignature generation in milliseconds.
+    pub presignature_timeout: u64,
+    /// Timeout for signature generation in milliseconds.
+    pub signature_timeout: u64,
+}
+
+impl Default for Config {
+    fn default() -> Self {
+        Self {
+            triple_timeout: min_to_ms(20),
```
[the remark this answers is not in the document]
ChaoticTempest: nope, this is just to showcase what we can do for the config side

chain-signatures/contract/src/config.rs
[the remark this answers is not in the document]
ChaoticTempest: yes, will make the configs consistent in a PR today or tomorrow

ChaoticTempest: I'm planning on adding dynamic configs so we don't have to change the contract just to update the config structure so there's little risk here. Nodes can have their own version of these configs they can override as well

chain-signatures/contract/src/errors.rs
ChaoticTempest: This is more for calling into residuals like `err?` so we can forward the error as a `MpcContractError`

chain-signatures/contract/src/lib.rs
ChaoticTempest: I wanted to keep all this info at the front and center so we don't miss out on changes

chain-signatures/contract/tests/tests.rs
ChaoticTempest: thanks for noticing this one. I made name changes on the function

chain-signatures/contract/tests/tests.rs
ChaoticTempest: yep, not trying to track conflicting proposals. Just going to generate a new id per proposal

chain-signatures/contract/src/config.rs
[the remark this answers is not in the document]
ppca: loll unfortunately this is what we have now in backend code

chain-signatures/contract/tests/tests.rs
ChaoticTempest: I'll add a status_msg contract to show that it does indeed change if we send in a different contract. I'm trying to figure out a way to safeguard against a case where we deploy unwanted contract. In that case, I think we should just validate the core interfaces to see that they still exist like:
- state
- sign
- respond

ppca (approved):
LGTM. 
Just wondering is there anyway we could check that the included wasm code is valid? Because even a benign partner could accidentally propose some bad code.
And what is our remedy if partner gets compromised and propose malicious contract code? But I guess we'd be the only official source to propose update and partners (assuming most are benign) will only vote for our proposed update.

chain-signatures/contract/src/update.rs
```diff
@@ -0,0 +1,101 @@
...
+    /// Vote for the update with the given id.
+    ///
+    /// Returns Some(votes) if the given [`UpdateId`] exists, otherwise None.
+    pub fn vote(&mut self, id: &UpdateId, voter: AccountId) -> Option<&HashSet<AccountId>> {
+        let votes = self.votes.get_mut(id)?;
+        votes.insert(voter);
+        Some(votes)
+    }
+
+    pub fn remove(&mut self, id: &UpdateId) -> Option<Vec<Update>> {
```
ppca, line 77: So if partner keeps proposing updates and no do_update is done, the updates map will keep piling up? Will that cause issue some day?

chain-signatures/contract/tests/tests.rs
ppca: +1 to validating interface. We should probably also check that contract is still upgradable, otherwise it's locked

chain-signatures/contract/src/update.rs
ChaoticTempest: yeah, we need to charge a deposit for propose_update

ChaoticTempest:
@ppca it's really hard to test valid upgrades. What I'm thinking is checking core interfaces like `state`, `sign` and `respond` to see if the contract got updated correctly but that would be at the contract level.

We can actually make voters pull down the update and verify locally it themselves to see if everything is running correctly. We probably need a CLI tool for this to run through the checks, but right now not a priority

ChaoticTempest:
Merging this one for now to let @ppca test the release candidate, will make a separate PR to address the issues noted here

chain-signatures/contract/src/lib.rs
ChaoticTempest: yeah we should flatten the errors out similar to workspace errors because we have too many similar sounding errors that can pop up for categories of errors. Let me see if I can whip up something real quick for this before release

chain-signatures/contract/tests/tests.rs
ChaoticTempest: the contract will always be upgradeable since there's no stopping the contract itself from calling deploy on itself even if the contract is locked

volovyks: yes, but if we break the function that is triggering this logic - we will loose the contract

PR #729, opened 2024-07-24
chore: additional changes to propose_update

ChaoticTempest:
Follow up to https://github.com/near/mpc/pull/714

- An attached deposit is now required on `propose_update` where the uploaded config or contract bytes is used to determine how much. This amount currently cannot be withdrawn, so it'll be trapped there until we make more changes to the contract. For us, this just means that there's a cost of about 8N per contract update to mainnet and not more than 0.1N for config updates
- Configs are now dynamic so we can make any arbitrary config update without changing the contract code at all
- `ProposedUpdates` now just uses one hashmap to save us the hassle of fetching both the votes and updates at the same time
- `migrate` function will be invoked after a deploy contract
- test case where deployed contract has a paniced/failed `migrate` function which effectively rolls back to the previous state. In this case the update fails and is evicted from being voted upon again


chain-signatures/contract/src/lib.rs
```diff
@@ -634,6 +640,22 @@ impl VersionedMpcContract {
...
+    /// This will be called internally by the contract to migrate the state when a new contract
+    /// is deployed. This function should be changed every time state is changed to do the proper
+    /// migrate flow.
+    ///
+    /// If nothing is changed, then this function will just return the current state. If it fails
+    /// to read the state, then it will return an error.
+    #[private]
+    #[init(ignore_state)]
+    #[handle_result]
+    pub fn migrate() -> Result<Self, MpcContractError> {
```
volovyks, line 652: We must keep the last deployed version somewhere to ensure migrations are successful.

chain-signatures/contract/src/lib.rs
```diff
@@ -772,12 +794,27 @@ impl VersionedMpcContract {
         }
     }
 
+    fn threshold(&self) -> Result<usize, VoteError> {
```
volovyks, line 797: VoteError looks strange here :/ We may need to reorg the error structure

chain-signatures/contract/src/update.rs
```diff
@@ -35,18 +44,29 @@ pub enum Update {
...
+    bytes_used: u128,
+}
+
 #[derive(Default, Debug, BorshSerialize, BorshDeserialize)]
 pub struct ProposedUpdates {
-    updates: HashMap<UpdateId, Vec<Update>>,
-    votes: HashMap<UpdateId, HashSet<AccountId>>,
-    next_id: u64,
+    entries: HashMap<UpdateId, UpdateEntry>,
+    id: UpdateId,
```
volovyks, line 57: Isn't next_id a better name?

chain-signatures/contract/src/update.rs
```diff
@@ -69,33 +93,60 @@ impl ProposedUpdates {
...
+                        "update_config".into(),
                         serde_json::to_vec(&(&config,)).unwrap(),
                         NearToken::from_near(0),
                         gas,
                     );
                 }
-                Update::Contract(code) => promise = promise.deploy_contract(code),
+                Update::Contract(code) => {
+                    // deploy contract then do a `migrate` call to migrate state.
+                    promise = promise.deploy_contract(code).function_call(
```
volovyks, line 121: Does this work as a batched transaction? Will deployment fail if the function call fails?

volovyks (approved):
LGTM, we will need to better test it. @ppca @ailisp I'm merging this to unblock other PRs, but please, take a look and comment.

chain-signatures/contract/tests/tests.rs
```diff
@@ -537,3 +563,63 @@ async fn test_propose_update_contract(contract: &Contract, accounts: &[Account])
...
+        .args_json(serde_json::json!({
+            "id": proposal_id,
+        }))
+        .transact()
+        .await
+        .unwrap();
+
+    dbg!(&execution);
+    let state: mpc_contract::ProtocolContractState = execution.json().unwrap();
+    dbg!(state);
```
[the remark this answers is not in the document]
ChaoticTempest: nope, `json().unwrap()` does it

chain-signatures/contract/tests/tests.rs
```diff
@@ -537,3 +563,63 @@ async fn test_propose_update_contract(contract: &Contract, accounts: &[Account])
...
+        if i == 1 {
+            dbg!(&execution);
+        }
+    }
+
+    // Try calling into state and see if it works after the contract updates with an invalid
+    // contract. It will fail in `migrate` so a state rollback on the contract code should have
+    // happened.
+    let execution = accounts[0]
+        .call(contract.id(), "state")
```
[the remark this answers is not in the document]
ChaoticTempest: because of the state rollback, the `FailureContract` does not get deployed, so we end up back at our original contract which has the `state` function

chain-signatures/contract/src/config.rs
```diff
@@ -1,23 +1,46 @@
...
-            triple_timeout: min_to_ms(20),
-            presignature_timeout: secs_to_ms(30),
-            signature_timeout: secs_to_ms(30),
-        }
+    #[serde(flatten)]
+    pub entries: HashMap<String, DynamicValue>,
+}
+
+impl Config {
+    pub fn get(&self, key: &str) -> Option<&serde_json::Value> {
```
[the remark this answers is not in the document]
ChaoticTempest: No, dynamic value is purely for internal representation. This is because `serde_json::Value` does not implement borsh serialization. Dynamic value is just a wrapper to facilitate our storage of the config in contract state

chain-signatures/contract/src/lib.rs
ChaoticTempest: yeah, I think we should have a list of structs when we're making these breaking changes

chain-signatures/contract/src/update.rs
ChaoticTempest: It's just an id generator. Calling it next_id meant that it could be sequential when it just generates you an arbitrary id

chain-signatures/contract/src/update.rs
ChaoticTempest: yeah, this ends up being a batch call where calling into `migrate` will roll back the whole entire promise with the deployed contract

PR #731, opened 2024-07-25
feat: use contract configs

ChaoticTempest:
This make use of contract configs in our node.

- `--override-config` parameter for partial overrides over the config in the contract
- we have less environment variables needed now, alongside less CLI arguments too


volovyks:
@ChaoticTempest I've fixed a conflict and contract tests, please, check the latest commits.
Part of the contract update test is commented. We are hitting transaction size.

chain-signatures/contract/tests/tests.rs
```diff
@@ -404,7 +404,7 @@ async fn test_contract_propose_update() {
     dbg!(contract.id());
 
     test_propose_update_config(&contract, &accounts).await;
-    test_propose_update_contract(&contract, &accounts).await;
+    // _test_propose_update_contract(&contract, &accounts).await;
```
volovyks, line 407: The contract is too big.

volovyks (approved):
LGTM!
Not sure what is going on with cargo check, it is all good on my machine and GCP VM.

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -241,6 +237,22 @@ impl MpcSignProtocol {
                 };
                 tracing::debug!(?contract_state);
 
+                // Sets the latest configurations from the contract:
```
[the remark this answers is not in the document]
ChaoticTempest: this is fetched every 1s alongside the contract state. I'm not sure what you mean by reboot then

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -241,6 +237,22 @@ impl MpcSignProtocol {
...
+                    &self.ctx.mpc_contract_id,
+                    &self.ctx.cfg,
+                )
+                .await
+                {
+                    Ok(config) => config,
+                    Err(e) => {
+                        tracing::error!("could not fetch contract's config: {e}");
+                        tokio::time::sleep(Duration::from_secs(1)).await;
+                        continue;
```
ailisp, line 252: Is one second too frequent, config is not very frequently changed. 

And above `rpc_client::fetch_mpc_config.await` blocks the main tokio thread, slow near rpc could slowdown protocol state handle(). I would suggest merge config with state call, doing it less frequently, or fetch it in a background tokio thread, without `await`, and `run` read from the fetched config written by the background thread, it could be 1s old but next MpcSign will get it

ailisp (approved):
Nice improvement! Although we didn't test the new config is applied in an integration test, I think the code looks right, config is refreshed and pass as function arguments instead of state to protocol modules, they'll always get latest config.

volovyks:
@ChaoticTempest I'm merging this to avoid conflicts. Let's address all the comments in a follow-up PR.

chain-signatures/node/src/protocol/mod.rs
ChaoticTempest: yeah 1s might be too frequent actually for configs that rarely if ever change. We can probably have a background task that fetches the config every 10 minutes or so as to not block this thread

PR #735, opened 2024-07-26
Flat contract errors

volovyks:


ChaoticTempest:
So, we're going to run into issues later with this when we get multiple errors being returned from the same function and they're not easily combined. The flat errors I was talking about is not where we have multiple enums, but rather a singular error type with an Error representation similar to [workspaces::ErrorRepr](https://github.com/near/near-workspaces-rs/blob/6c16f12f84e5c43b03f9f757ae2f11d1c6ebf32e/workspaces/src/error/mod.rs#L36). That way we don't need to think about combining errors and just simply give the error code with custom message or wrap from another error

volovyks:
@gagdiez can you take a look at this, please? Not sure what is best. @ChaoticTempest design is more flexible, but it is not visible what category of errors the function returns.

volovyks:
Closing in favour of https://github.com/near/mpc/pull/757

PR #737, opened 2024-07-27
chore: made ProposedUpdates use IterableMap underneath due contracts taking too much space

ChaoticTempest:
- Reorganized tests in contracts/tests
- Added test for proposing multiple contracts such that we don't run into the case where contracts are being loaded at every state load into memory; and then costing us a lot of gas per function call.
- ProposedUpdates now uses `IterableMap` internally due to the above.


ChaoticTempest:
@ppca this is a breaking change for RC1

PR #752, opened 2024-07-29
feat: make PresignatureId be generated and reproducible from triple ids

ChaoticTempest:
This is better to have now than later, as we won't break the presignature id guarantees later if we add it later.

This will allow us to validate `Presignature` generation messages a lit bit more easily by verifying that the presignature id is correct in the future

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -10,6 +10,8 @@ use chrono::Utc;
 use crypto_shared::PublicKey;
 use k256::Secp256k1;
 use mpc_contract::config::ProtocolConfig;
+use rand::rngs::StdRng;
+use rand::{Rng as _, SeedableRng as _};
 use std::collections::hash_map::Entry;
 use std::collections::{HashMap, HashSet, VecDeque};
 use std::time::{Duration, Instant};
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -516,3 +518,15 @@ impl PresignatureManager {
...
 }
+
+pub fn hash_as_id(triple0: TripleId, triple1: TripleId) -> PresignatureId {
+    use sha3::{Digest, Sha3_256};
+    let mut hasher = Sha3_256::new();
+    hasher.update(triple0.to_le_bytes());
+    hasher.update(triple1.to_le_bytes());
+    let seed: [u8; 32] = hasher.finalize().into();
+    let mut rng = StdRng::from_seed(seed);
+    let id = rng.gen::<u64>();
```
ailisp, line 529: Why not just pick first 8 byte of the `[u8; 32]` as u64 id?

ailisp (approved):
Nice enhancement!

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: yeah that also works. Either could work, but I guess that one doesn't rely on RNG. Will change it to not use RNG

volovyks (approved):
Nice.
test_multichain_reshare_with_lake_congestion test is unstable :/ cc @ailisp I do not think that this is fault of the test. Probably we need to fix it on the node some day.

PR #754, opened 2024-07-29
A collection of small audit fixes

DavidM-D:


chain-signatures/node/src/protocol/triple.rs
```diff
@@ -309,10 +310,16 @@ impl TripleManager {
...
             self.gc.insert(id1, Instant::now());
-            Ok((
-                self.triples.remove(&id0).unwrap(),
-                self.triples.remove(&id1).unwrap(),
-            ))
+
+            let triple_0 = self
+                .triples
+                .remove(&id0)
+                .ok_or(GenerationError::TripleIsMissing(id0))?;
```
ChaoticTempest, line 317: It doesn't hurt to add the `ok_or`s but we already guarantee that they exist with the previous `self.triples.contains_key({id0, id1})`

ChaoticTempest (approved):

PR #757, opened 2024-07-29
Flat error handling

volovyks:
Flat and opaque way to handle errors.
@ChaoticTempest This approach allowed us to move the errors between categories and remove all error duplications. Now I like it more :)


volovyks:
@ChaoticTempest I've fixed the tests

chain-signatures/contract/src/errors/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    #[error("Account to join is not in the candidate set.")]
+    JoinNotCandidate,
+    #[error("Number of participants cannot go below threshold.")]
+    ParticipantsBelowThreshold,
+    #[error("Update not found.")]
+    UpdateNotFound,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone, thiserror::Error)]
+pub enum Common {
```
ChaoticTempest, line 61: Now that we have this pattern, we shouldn't be making a common kind. Instead we should separate out all of these into their own category such as:

```
enum ProtocolError {
    StateNotRunning,
    ...
}

enum InvalidState {
    RequestNotFound,
    
}

enum InvalidParameters {
    InsufficientDeposit,
    InsufficientGas,
    ...
}

```

volovyks: @ChaoticTempest This is opinionated and generates unnecessary code. Error representation for the user will not change, but for us, it can add some confusion. For example, `RequestNotFound` is `InvalidState` and `InvalidParameters` error at the same time. Same about `EpochMismatch`.

ChaoticTempest: hmm, I think `Common` is too ambiguous of name since that doesn't signify how that error could have happened. I rather us base it on opinionated items than make it ambiguous.

Then I guess `RequestNotFound` can go in `InvalidParameters` since I don't see it being using anywhere else besides for invalid parameters. `EpochMismatch` is more `InvalidState` though

volovyks: Done

ChaoticTempest (approved):

volovyks:
Let's do the follow up for RUSTSEC-2024-0357

PR #759, opened 2024-07-30
fix: make override default impl return empty map and use u32 instead of usize

ChaoticTempest:
fixes:
- make override default have empty map instead of null
- make all usize in contract/ code use u32 instead due to architectural differences


ppca (approved):
LGTM

ppca (approved):

PR #760, opened 2024-07-30
fix: make polling config less frequent

ChaoticTempest:
This makes polling config less frequent as configs very rarely change if ever for now.

This updates the poll time from 1 second to 5 minutes


ailisp (approved):
Looks good to me. One problem to keep in mind is the configuration difference between node could exist for up to 5 minutes. As my understanding current protocol config different won't cause problem, but future added could be, and please double think of current protocol config as well in case there is problematic config that I'm not aware of.

volovyks (approved):
It can lead to inconsistencies. But that is nothing compared to the manual update.

PR #761, opened 2024-07-30
Make the signature cost public

DavidM-D:


volovyks (commented):
The price will not be guaranteed to stay the same during the `sign()` call.

volovyks:
Closing in favor of https://github.com/near/mpc/pull/782

PR #765, opened 2024-07-30
Load test fixes

volovyks:
@ppca Fixed acc name and log string. We need to deploy this code to our test envs to make it work. Overall load tests were functioning, I've added a few comments for better understanding. The error rate is over the roof because our dev env is malfunctioning.


chain-signatures/contract/src/lib.rs
```diff
@@ -728,7 +728,10 @@ impl VersionedMpcContract {
     ) -> Result<SignatureResponse, MpcContractError> {
         match self {
             Self::V0(_) => match signature {
-                SignatureResult::Ok(signature) => Ok(signature),
+                SignatureResult::Ok(signature) => {
+                    log!("Signature is ready.");
```
volovyks, line 732: I expect this log to appear in sign() call logs

ppca (approved):

PR #768, opened 2024-07-30
Payments with balance

### LINKED ISSUE #23: Alternative sign() payment design
Currently, we have the `#[payable]` `sign()` function with a dynamic pricing model.
Downsides:
- Unpredictable price, users do not know the exact price of a signature.
- Each change of the minimum signature price is a breaking change.

Alternative design:
- `sign()` always requires 1 yN of deposit
- users must call `fund()` and attach a deposit of their choice
- Users are charged for each successful `sign()` request from their contract balance

The alternative design allows both static and dynamic `sign()` request price models. I'm leaning toward a static one.

volovyks:


chain-signatures/contract/src/lib.rs
```diff
@@ -254,6 +277,11 @@ impl VersionedMpcContract {
 
             match self {
                 Self::V0(mpc_contract) => {
+                    // withdraw the signature fee
+                    mpc_contract
+                        .balances
+                        .withdraw(&request.account_id, request.signature_fee);
```
gagdiez, line 283: @volovyks the recommended pattern for handling balances is: 

- withdraw on request
- refund on error

Otherwise, you could get a lot of requests, which you approve since the user has the balance, and then discover during the response that the user actually has no money to pay for them all

So in this case, we would want to use `balances.withdraw` on `sign`, and `return_signature` would refund the user if they never got the signature because of a response timeout

volovyks: That is true. We can fail in multiple places, so this design seemed simpler and more robust.

PR #779, opened 2024-07-31
fix: signature request order and retrying

ChaoticTempest:
This fixes:
- signature retrying constantly even after request has fully timeout on contract. This adds in total generation timeout to alleviate that.
- signature request ordering is very arbitrary. Could likely go into a case where we don't pick the signature request and it starves. This fixes that by adding insertion order into the requests as one of the metrics for choice.


volovyks (approved):
LGTM

Issue #708, opened 2024-07-22
Refund if sign() failed


volovyks:
We need to refund the user's deposit if the sign() call fails for any reason. But let's ensure it is impossible to drain funds from our contract with bad requests.
Proposed by @gagdiez [here](https://github.com/near/mpc/pull/706#discussion_r1686465787).

ailisp:
There are three cases to consider:
1. when the `sign` call fails (not the cross contract call invoked by `sign`), near will automatically return attached deposit back to the caller: https://docs.near.org/build/smart-contracts/anatomy/environment#attached-deposit
2. when the `sign` call succeed, we should refund the difference, this is the [proposal](https://github.com/near/mpc/pull/706#discussion_r1686465787) by @gagdiez 
3. when the `sign` call success, but the cross contract call invoked by `sign`, i.e., `sign_helper`, `clear_state_on_finish`, `return_signature_on_finish` fails, we should refund the remaining over required_deposit

volovyks:
@ailisp 
1. Ok
2. Agree
3. Why not all? It is not the user's fault that one of our functions failed.

volovyks:
Blocked by https://github.com/near/mpc/issues/767

volovyks:
We will push back with #767 for now.

PR #781, opened 2024-07-31
refund after sign

ailisp:
Refund all deposit if sign fails.
- If sign call itself fails (does not include the cross contract call invoked by sign, for example incorrect version), near runtime will not take required deposit, does not need to refund
- sign will refund more than required deposit if sign it self fails, then if sign invoked cross contract call fails, it further refund required deposit. As a result all deposit is refunded

Refund more than required deposit if sign success.

Address #708


chain-signatures/contract/src/lib.rs
```diff
@@ -718,15 +731,28 @@ impl VersionedMpcContract {
     pub fn clear_state_on_finish(
         &mut self,
         request: SignatureRequest,
+        refund_to_on_fail: AccountId,
```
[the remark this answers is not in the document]
ailisp: Yes, make it only refund once now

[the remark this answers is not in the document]
ailisp: SignatureRequest also used in protocol, in protocol SignatureManager.poke, the refund and requester info isn't relevant in that context, therefore I'll make another ContractSignatureRequest struct enclosing the original SignatureRequest, use specifically for contract call

chain-signatures/contract/src/lib.rs
```diff
@@ -696,7 +692,7 @@ impl VersionedMpcContract {
                 // by it won't execute.
                 let yield_promise = env::promise_yield_create(
                     "clear_state_on_finish",
-                    &serde_json::to_vec(&(&contract_signature_request)).unwrap(),
+                    &serde_json::to_vec(&(&contract_signature_request,)).unwrap(),
```
ailisp, line 695: Can't believe missing this `,` took me an afternoon to fix 😿

chain-signatures/contract/src/lib.rs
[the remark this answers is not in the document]
ailisp: Done

chain-signatures/contract/src/lib.rs
```diff
@@ -677,12 +683,16 @@ impl VersionedMpcContract {
     }
 
     #[private]
-    pub fn sign_helper(&mut self, request: SignatureRequest) {
+    pub fn sign_helper(&mut self, contract_signature_request: ContractSignatureRequest) {
         match self {
             Self::V0(mpc_contract) => {
+                // refund must happen in clear_state_on_finish, because regardless of this success or fail
```
volovyks, line 689: Should we move this comment to where the refund is actually happening?

volovyks (dismissed):

chain-signatures/contract/src/lib.rs
ailisp: OK

chain-signatures/contract/tests/sign.rs
```diff
@@ -57,6 +59,124 @@ async fn test_contract_sign_request() -> anyhow::Result<()> {
...
+
+    let new_balance = alice.view_account().await?.balance;
+    println!(
+        "{} {} {}",
+        balance.as_millinear(),
+        new_balance.as_millinear(),
+        contract.view_account().await?.balance
+    );
+    assert!(
+        balance.as_millinear() - new_balance.as_millinear() < 10,
```
volovyks, line 173: Sorry for the late comment. Can we also ensure that our balance is not lower than expected?

volovyks: I'm talking about mpc-contract balance

ailisp: This is implicitly ensured, i tested manually an intentional over refund, which cause this line underflow and fail. I can make the check explicit

ailisp:
@volovyks ptal

ppca (approved):
LGTM

PR #789, opened 2024-07-31
fix: make indexer behind and running threshold parametrizable

ChaoticTempest:
Feels like these values should be changeable in general. Also for testing purposes we want probably want them to be lower


ppca (dismissed):

volovyks:
Will we use default values for testnet/mainnet? I do not want them to rebuild everything again.

ChaoticTempest:
Also made default `BEHIND_THRESHOLD` to be 3 mins because indexer can be slower on mainnet and we don't want to mark these nodes as unstable

volovyks (approved):

PR #790, opened 2024-07-31
test: multi signature production

ChaoticTempest:
Adds in test for doing signature production chaotically throughout a whole entire test

- Also reduces the number of debug logs that were redundant where other logs already log that info

### DIFF
--- a/integration-tests/chain-signatures/tests/actions/wait_for.rs
+++ b/integration-tests/chain-signatures/tests/actions/wait_for.rs
@@ -311,6 +311,61 @@ pub async fn signature_responded(
     }
 }
 
+pub async fn signatures_responded(
+    ctx: &MultichainTestContext<'_>,
+    tx_hashs: &[CryptoHash],
+) -> Result<Vec<FullSignature<Secp256k1>>, WaitForError> {
+    let is_tx_ready = |tx_hash: CryptoHash| async move {
+        let outcome_view = match ctx
+            .jsonrpc_client
+            .call(RpcTransactionStatusRequest {
+                transaction_info: TransactionInfo::TransactionId {
+                    tx_hash,
+                    sender_account_id: ctx.nodes.ctx().mpc_contract.id().clone(),
+                },
+                wait_until: near_primitives::views::TxExecutionStatus::Final,
+            })
+            .await
+        {
+            Err(error) => return Err(WaitForError::JsonRpc(format!("{error:?}"))),
+            Ok(outcome_view) => outcome_view,
+        };
+
+        let Some(outcome) = outcome_view.final_execution_outcome else {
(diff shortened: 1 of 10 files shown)


integration-tests/chain-signatures/tests/cases/mod.rs
```diff
@@ -99,6 +100,41 @@ async fn test_signature_basic() -> anyhow::Result<()> {
...
+                let sig_amount = rng.gen_range(1, 3);
+                tracing::info!(i, sig_amount, "Producing signature");
+
+                let (_payload_hashes, tx_hashes): (Vec<_>, Vec<_>) =
+                    actions::request_sign_multiple(&ctx, &signer, sig_amount)
+                        .await?
+                        .into_iter()
+                        .unzip();
+
+                match wait_for::signatures_responded(&ctx, &tx_hashes).await {
```
volovyks, line 126: From what I see in `signatures_responded()` fn there is no guarantee that it will return all signatures. Some may be in progress. Should we check the length here? Or make sure we have N signatures in the function?

integration-tests/chain-signatures/tests/actions/wait_for.rs
```diff
@@ -311,6 +311,61 @@ pub async fn signature_responded(
...
+    };
+    let mut result = Vec::new();
+    for _ in 0..8 {
+        for hash in tx_hashs {
+            match is_tx_ready(*hash).await {
+                Ok(Outcome::Signature(signature)) => result.push(signature),
+                Ok(Outcome::Failed(err)) => {
+                    return Err(WaitForError::Signature(SignatureError::Failed(err)))
+                }
+                Err(_) => continue,
```
ailisp, line 361: This may result in one result pushed more than once to result, I suggest use a Map and skip checking already in Map hashs.

integration-tests/chain-signatures/tests/cases/nightly.rs
```diff
@@ -11,41 +13,36 @@ async fn test_nightly_signature_production() -> anyhow::Result<()> {
...
+
+                let (_payload_hashes, tx_hashes): (Vec<_>, Vec<_>) =
+                    actions::request_sign_multiple(&ctx, &signer, sig_amount)
+                        .await?
+                        .into_iter()
+                        .unzip();
+
+                match wait_for::signatures_responded(&ctx, &tx_hashes).await {
+                    Ok(_signatures) => tracing::info!(?tx_hashes, "got signatures"),
+                    Err(err) => tracing::error!("unable to produce signatures in time: {err:?}"),
```
ailisp, line 45: Would be helpful to collect some metrics: success percentage, average signature time. and we can compare daily result to know if our system is improved or hitting some regression

PR #792, opened 2024-08-01
fix: give indexer some time to catch up before failing on restart

ppca:


chain-signatures/node/src/indexer.rs
```diff
@@ -324,7 +328,7 @@ pub fn run(
                 anyhow::Ok(lake)
             }) else {
                 tracing::error!(?options, "indexer failed to build");
-                backoff(i);
+                backoff(i, 1, 120);
```
ailisp, line 331: Seems i will never decrease, and eventually i will become `max`  if indexer keep working. Do we need to restore it to a small number on a not failing case?

chain-signatures/contract/src/lib.rs
```diff
@@ -584,6 +584,7 @@ impl VersionedMpcContract {
 #[near_bindgen]
 impl VersionedMpcContract {
     #[handle_result]
+    #[init(ignore_state)]
```
volovyks, line 587: Why do we need this change? Not sure if it's safe.

ppca: ah sorry. shouldn't have pushed this.

chain-signatures/node/src/indexer.rs
```diff
@@ -349,6 +353,10 @@ pub fn run(
                 rt.spawn(async move { lake.run_with_context_async(handle_block, &context).await })
             };
             let outcome = rt.block_on(async {
+                if i > 0 {
+                    // give it some time to catch up
+                    backoff(i, 10, 300);
+                }
```
ChaoticTempest, line 359: we could've just had a flat `sleep(5mins)` here instead of modifying `backoff`

ChaoticTempest (approved):

Issue #764, opened 2024-07-30
Test all contract APIs (Unit tests)


volovyks:

ailisp:
init, sign, respond and vote to update part of APIs are tested. The missing part is join and vote_ join/reshared/leave/pk.

volovyks:
@ailisp I would prioritize user-facing functions. Voting and resharing are important too.
```rust
fn sign() // tested
fn public_key()
fn derived_public_key()
fn latest_key_version()
fn experimantal_signature_deposit()
```

PR #794, opened 2024-08-02
add more contract api tests

ailisp:
#764

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -135,7 +135,7 @@ impl VersionedMpcContract {
         }
         // Check deposit
         let deposit = env::attached_deposit();
-        let required_deposit = self.experimantal_signature_deposit();
+        let required_deposit = self.experimental_signature_deposit();
         if deposit.as_yoctonear() < required_deposit {
             return Err(InvalidParameters::InsufficientDeposit.message(format!(
                 "Attached {}, Required {}",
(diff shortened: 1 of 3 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -215,10 +215,10 @@ impl VersionedMpcContract {
         0
     }
 
-    /// This experimantal function calculates the fee for a signature request.
+    /// This experimental function calculates the fee for a signature request.
     /// The fee is volatile and depends on the number of pending requests.
     /// If used on a client side, it can give outdate results.
-    pub fn experimantal_signature_deposit(&self) -> u128 {
```
volovyks, line 221: Thank you! I think I've made that mistake in docs as well.

chain-signatures/contract/tests/user_views.rs
```diff
@@ -0,0 +1,99 @@
...
+
+    let key: String = contract.view("public_key").await.unwrap().json().unwrap();
+    println!("{:?}", key);
+    let pk = PublicKey::from_str(&key)?;
+    assert_eq!(pk.curve_type(), CurveType::SECP256K1);
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_derived_public_key() -> anyhow::Result<()> {
```
volovyks, line 35: We will need to execute actual derivation logic in such a test. Not a blocker now.

chain-signatures/contract/tests/vote.rs
```diff
@@ -0,0 +1,351 @@
...
+    Ok(())
+}
+
+#[tokio::test]
+async fn test_vote_pk() -> anyhow::Result<()> {
+    let (_, contract, accounts, _) = init_env().await;
+
+    let key: String = contract.view("public_key").await.unwrap().json().unwrap();
+
+    let execution = accounts[2]
```
volovyks, line 207: We should check that random acc can not vote for pk

volovyks (approved):
Great work!

chain-signatures/contract/tests/user_views.rs
ppca: agree. Probably good to be able to fix the pk in init_env() so we can test the value of the derived pk is correct.

ppca (approved):

volovyks:
@ailisp merging, lets address all the comments if a follow up PR

PR #797, opened 2024-08-02
Logs, log levels

volovyks:
Logs are opinionated. I want us to start discussing how, where, and what we are logging.

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -110,6 +110,10 @@ pub struct Indexer {
 
 impl Indexer {
     fn new(latest_block_height: LatestBlockHeight, options: &Options) -> Self {
+        tracing::info!(
+            "creating new indexer, latest block height: {}",
+            latest_block_height.block_height
+        );
         Self {
             latest_block_height: Arc::new(RwLock::new(latest_block_height)),
             last_updated_timestamp: Arc::new(RwLock::new(Instant::now())),
@@ -143,6 +147,7 @@ impl Indexer {
         block_height: BlockHeight,
         gcp: &GcpService,
     ) -> Result<(), DatastoreStorageError> {
+        tracing::trace!(block_height, "update_block_height");
         *self.last_updated_timestamp.write().await = Instant::now();
         self.latest_block_height
             .write()
@@ -166,10 +171,11 @@ async fn handle_block(
     mut block: near_lake_primitives::block::Block,
     ctx: &Context,
 ) -> anyhow::Result<()> {
-    tracing::debug!(block_height = block.block_height(), "handle_block");
+    tracing::trace!(block_height = block.block_height(), "handle_block");
     let mut pending_requests = Vec::new();
     for action in block.actions().cloned().collect::<Vec<_>>() {
         if action.receiver_id() == ctx.mpc_contract_id {
+            tracing::trace!("got action targeting {}", ctx.mpc_contract_id);
             let Some(receipt) = block.receipt_by_id(&action.receipt_id()) else {
                 let err = format!(
                     "indexer unable to find block for receipt_id={}",
@@ -185,6 +191,7 @@ async fn handle_block(
                 continue;
             };
             if function_call.method_name() == "sign" {
+                tracing::trace!("found `sign` function call");
                 let arguments =
                     match serde_json::from_slice::<'_, SignArguments>(function_call.args()) {
                         Ok(arguments) => arguments,
@@ -326,14 +333,14 @@ pub fn run(
         let mut i = 0;
         loop {
             if i > 0 {
-                tracing::info!("restarting indexer after failure: restart count={i}");
+                tracing::warn!("restarting indexer after failure: restart count={i}");
             }
             i += 1;
 
             let Ok(lake) = rt.block_on(async {
                 let latest = context.indexer.latest_block_height().await;
                 if i > 0 {
-                    tracing::info!("indexer latest height {latest}, restart count={i}");
+                    tracing::warn!("indexer latest height {latest}, restart count={i}");
                 }
                 let mut lake_builder = LakeBuilder::default()
                     .s3_bucket_name(&options.s3_bucket)
@@ -365,6 +372,7 @@ pub fn run(
             let outcome = rt.block_on(async {
                 if i > 0 {
                     // give it some time to catch up
+                    tracing::trace!("giving indexer some time to catch up");
                     backoff(i, 10, 300);
                 }
                 // while running, we will keep the task spinning, and check every so often if
@@ -378,6 +386,7 @@ pub fn run(
 
                 // Abort the indexer task if it's still running.
                 if !join_handle.is_finished() {
+                    tracing::trace!("aborting indexer task");
                     join_handle.abort();
                 }
 
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -165,6 +165,7 @@ async fn handle_block(
     mut block: near_lake_primitives::block::Block,
     ctx: &Context,
 ) -> anyhow::Result<()> {
+    tracing::debug!(block_height = block.block_height(), "handle_block");
     let mut pending_requests = Vec::new();
     for action in block.actions().cloned().collect::<Vec<_>>() {
         if action.receiver_id() == ctx.mpc_contract_id {
@@ -205,10 +206,13 @@ async fn handle_block(
                     continue;
                 };
 
-                let Ok(entropy) = serde_json::from_str::<'_, [u8; 32]>(&receipt.logs()[1]) else {
+                let entropy_log_index = 1;
+                let Ok(entropy) =
+                    serde_json::from_str::<'_, [u8; 32]>(&receipt.logs()[entropy_log_index])
+                else {
                     tracing::warn!(
                         "`sign` did not produce entropy correctly: {:?}",
-                        receipt.logs()[0]
+                        receipt.logs()[entropy_log_index]
                     );
                     continue;
                 };
@@ -260,9 +264,6 @@ async fn handle_block(
     }
     drop(queue);
 
-    if block.block_height() % 1000 == 0 {
-        tracing::info!(block_height = block.block_height(), "indexed block");
-    }
     Ok(())
 }
 
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -265,6 +265,15 @@ async fn handle_block(
     }
     drop(queue);
 
+    let log_indexing_interval = 1000;
+    if block.block_height() % log_indexing_interval == 0 {
+        tracing::info!(
+            "indexed another {} blocks, latest: {}",
+            log_indexing_interval,
+            block.block_height()
+        );
+    }
+
     Ok(())
 }
 
--- a/chain-signatures/node/src/mesh/connection.rs
+++ b/chain-signatures/node/src/mesh/connection.rs
@@ -74,10 +74,6 @@ impl Pool {
 
         let mut active = self.current_active.write().await;
         *active = Some((participants.clone(), Instant::now()));
-        tracing::debug!(
-            "active participant account id's {:?}",
-            participants.account_ids()
-        );
         participants
     }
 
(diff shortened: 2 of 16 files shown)


chain-signatures/node/src/indexer.rs
```diff
@@ -260,9 +264,6 @@ async fn handle_block(
     }
     drop(queue);
 
-    if block.block_height() % 1000 == 0 {
-        tracing::info!(block_height = block.block_height(), "indexed block");
-    }
```
ChaoticTempest, line 265: we should keep this as it was nice to see if indexer was still running or not. It doesn't spam the logs either

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -35,6 +35,7 @@ impl Pool {
             }
         }
 
+        tracing::debug!("pinging participants");
```
ChaoticTempest, line 38: these stop debug logs should not be necessary. Pinging connections can never fail so we'll always call into the following debug log

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -84,6 +89,8 @@ impl Pool {
             }
         }
 
+        tracing::debug!("pinging potential participants");
```
ChaoticTempest, line 92: same with this one

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -143,6 +154,10 @@ impl Pool {
 
     async fn set_potential_participants(&self, participants: &Participants) {
         *self.potential_connections.write().await = participants.clone();
+        tracing::debug!(
+            "Pool set potential participants to {:?}",
+            self.potential_connections.read().await.keys_vec()
+        );
     }
```
ChaoticTempest, line 161: I'd vote for getting rid of these `setting the participants` logs as the contract state already logs the participants

chain-signatures/node/src/rpc_client.rs
```diff
@@ -12,6 +12,7 @@ pub async fn fetch_mpc_contract_state(
 ) -> anyhow::Result<ProtocolState> {
     let protocol_state: mpc_contract::ProtocolContractState =
         rpc_client.view(mpc_contract_id, "state").await?.json()?;
+    tracing::debug!(?protocol_state, "fetched protocol state");
```
ChaoticTempest, line 15: we already have a debug for contract state

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -108,6 +115,10 @@ impl Pool {
 
         let mut potential_active = self.potential_active.write().await;
         *potential_active = Some((participants.clone(), Instant::now()));
+        tracing::debug!(
+            "potential participant account id's {:?}",
+            participants.account_ids()
+        );
```
ChaoticTempest, line 121: I'd say we should merge logging the potential participants and active participants into one call in mesh, because this ping call happens every 300ms. It's going to be really hard to find what we want

chain-signatures/node/src/rpc_client.rs
ppca: +1

chain-signatures/node/src/indexer.rs
ppca: ok to keep this one. It's helpful for us to see what are the logs around the time block height increase stopped.

ppca (commented):
I'd say we should bring "received a new message" from tracing back to debug, but that is indeed spammy. But we kinda need a way to figure out if a node is able to receive messages and was receiving messages.

chain-signatures/node/src/indexer.rs
volovyks: I'm not fond of such a pattern, since it is inconsistent. What if 1 block dropped? What if we need to know about every block we are receiving? I've changed the message to eliminate those.

And I've added a debug!() log that is printing all the new blocks.

volovyks:
@ppca @ChaoticTempest Overall, I think we should be looser with debug logs. They should give as full understanding of what is going on in our system, not just the "highlights". Info is for highlights, it should work fine once we have log filtering.

volovyks:
@ppca @ChaoticTempest I've added more logs, and removed or moved some of them. Changed log levels, and now we should be able to use them in GCP.

ppca:
Adding logs look good to me. But we still need to enable toggling for this to work in GCP?

volovyks:
@ppca are you talking about this https://github.com/near/mpc/pull/803 ?

chain-signatures/node/src/mesh/connection.rs
ChaoticTempest: we can get rid of these now since https://github.com/near/mpc/pull/800 already does what I mentioned

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -57,6 +57,7 @@ impl Mesh {
                 stable.insert(participant, info.clone());
             }
         }
+        tracing::info!(?stable, "stable participants");
```
ChaoticTempest, line 60: This should be placed at call site instead of inside the function otherwise multiple calls to it will spam the logs and we should know where exactly the context of stable is

chain-signatures/node/src/protocol/message.rs
```diff
@@ -218,6 +218,7 @@ impl MessageHandler for ResharingState {
         let q = queue.resharing_bins.entry(self.old_epoch).or_default();
         let mut protocol = self.protocol.write().await;
         while let Some(msg) = q.pop_front() {
+            tracing::trace!("handling new resharing message");
```
ChaoticTempest, line 221: Instead of having a new log call per resharing message, we should just log how many resharing messages we have

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -166,9 +166,22 @@ impl MpcSignProtocol {
...
+        tracing::info!(
+            "initializing protocol with parameters:
+            my_address: {my_address},
+            mpc_contract_id: {mpc_contract_id},
+            account_id: {account_id},
+            rpc_client_url: {rpc_url},
+            signer_account_id: {signer_account_id},
+            cfg: {:?}",
+            cfg
+        );
```
ChaoticTempest, line 181: Just use the tracing formatter itself otherwise this will probably have weird formatting on the string with all these spaces inside the string:
```suggestion
        tracing::info!(
            my_address,
            mpc_contract_id,
            account_id,
            rpc_url,
            signer_account_id,
            ?cfg,
            "initializing protocol with parameters"
        );
```

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -304,7 +315,7 @@ impl MpcSignProtocol {
                 let from_state = format!("{state}");
                 state = match state.advance(&mut self, contract_state).await {
                     Ok(state) => {
-                        tracing::debug!("advance ok: {from_state} => {state}");
+                        tracing::trace!("advance ok: {from_state} => {state}");
```
ChaoticTempest, line 318: not sure if @ppca will like this because it might not be visible that we're advancing states in our normal debug logs

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -157,8 +157,13 @@ impl PresignatureManager {
     }
 
     pub fn garbage_collect(&mut self, cfg: &ProtocolConfig) {
+        let before = self.gc.len();
         self.gc
             .retain(|_, instant| instant.elapsed() < Duration::from_millis(cfg.garbage_timeout));
+        let removed = before.saturating_sub(self.gc.len());
+        if removed > 0 {
+            tracing::warn!("garbage collected {} presignatures", removed);
```
ChaoticTempest, line 165: I wouldn't say this is a warning though, since this is purely informational where warning should be reserved for logs that should close to never happen. So it should be `info` instead.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -118,18 +118,21 @@ impl SignQueue {
...
                 let proposer_requests = self.requests.entry(proposer).or_default();
                 proposer_requests.insert(request.receipt_id, request);
-                crate::metrics::NUM_SIGN_REQUESTS_MINE
-                    .with_label_values(&[my_account_id.as_str()])
-                    .inc();
+                if is_mine {
+                    crate::metrics::NUM_SIGN_REQUESTS_MINE
+                        .with_label_values(&[my_account_id.as_str()])
+                        .inc();
+                }
```
ChaoticTempest, line 135: Not sure if this is correct. @ppca is this the right metric for MINE sign requests or does this include requests that we are a part of the signer subset?

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -199,8 +202,9 @@ impl SignatureGenerator {
 
     pub fn poke(&mut self) -> Result<Action<FullSignature<Secp256k1>>, ProtocolError> {
         if self.sign_request_timestamp.elapsed() > self.timeout_total {
+            tracing::warn!("signature protocol timed out completely");
             return Err(ProtocolError::Other(
-                anyhow::anyhow!("signature protocol timeout completely").into(),
+                anyhow::anyhow!("signature protocol timed out completely").into(),
```
ChaoticTempest, line 207: I wouldn't rename it `timed out` because it's much easier to search for `timeout` even though `timed out` is more grammatically correct

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -737,9 +741,17 @@ impl SignatureManager {
 
     /// Garbage collect all the completed signatures.
     pub fn garbage_collect(&mut self, cfg: &ProtocolConfig) {
+        let before = self.completed.len();
         self.completed.retain(|_, timestamp| {
             timestamp.elapsed() < Duration::from_millis(cfg.signature.garbage_timeout)
         });
+        let garbage_collected = before.saturating_sub(self.completed.len());
+        if garbage_collected > 0 {
+            tracing::warn!(
```
ChaoticTempest, line 750: same here with this garbage collection

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -190,9 +190,14 @@ impl TripleManager {
 
     /// Clears an entry from failed triples if that triple protocol was created more than 2 hrs ago
     pub fn garbage_collect(&mut self, cfg: &ProtocolConfig) {
+        let before = self.gc.len();
         self.gc.retain(|_, timestamp| {
             timestamp.elapsed() < Duration::from_millis(cfg.garbage_timeout)
         });
+        let garbage_collected = before.saturating_sub(self.gc.len());
+        if garbage_collected > 0 {
+            tracing::warn!("garbage collected {} triples", garbage_collected);
```
ChaoticTempest, line 199:
```suggestion
            tracing::info!("garbage collected {} triples", garbage_collected);
```

chain-signatures/node/src/protocol/mod.rs
volovyks: I've made it trace because it is too spammy. We should fix it later with logging spans.

chain-signatures/node/src/protocol/presignature.rs
volovyks: I think we should minimize the number of garbage collections since it is a waste and an indicator that our system is not efficient.

chain-signatures/node/src/protocol/signature.rs
volovyks: This is an intentional change. I believe that it is more important now to have a metric for the proposer of the signature. But we can have both. Also, "MINE" semantics is closer to me=proposer.

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: I think you're misunderstanding what garbage collection is. Garbage collections happens no matter what, as in every single triple, presignature and signature will eventually be garbage collected and this will always be called. It is our way to keep tally of what IDs to keep around until messages stop from other protocol that linger behind or are delayed due to restarts or other misc behaviors

volovyks: kk, I got it
what I meant is that we should keep track of gcd triples, pre-signatures that were not used for any reason

ChaoticTempest (approved):

PR #801, opened 2024-08-03
feat: improve triple generation speed by spinning up as a tokio::task

ChaoticTempest:
We've been manually managing triple protocols via configs like `max_concurrent_generation` in our main runtime thread which is not quite efficient. By spawning them as tokio tasks, we should see a good speedup. We also no longer need to manage the amount of work we need ourselves, since the Tokio runtime has features like spinning up more threads to further accommodate the workload that we push to it, with work stealing such that no task takes too long.

From what I've seen our average elapsed=35s of triple completion went down to 15-25s averages with this change.

This is however a potential instability to add while we are releasing mainnet, so should only be merged if we have high confidence in this so please test this for yourself before pushing this in


ailisp (approved):
Great optimization! I think it looks safe, but it would be nice to move the tokio message sending code to our `impl cait_sith::Protocol` to replace cait-sith's `impl<'a, T> Protocol for ProtocolExecutor<'a, T> {`. If so we can keep our protocol mostly unchanged with the speed up implemented in this PR

ChaoticTempest:
I'll be closing this one for now as some more testing for this needs to be done

PR #804, opened 2024-08-07
metrics: add failure counter for triple/presig/signature

ppca:
failure counters

### DIFF
--- a/chain-signatures/node/src/metrics.rs
+++ b/chain-signatures/node/src/metrics.rs
@@ -335,6 +335,51 @@ pub(crate) static FAILED_SEND_ENCRYPTED_LATENCY: Lazy<HistogramVec> = Lazy::new(
     .unwrap()
 });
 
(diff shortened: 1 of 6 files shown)


ChaoticTempest (approved):

chain-signatures/node/src/metrics.rs
```diff
@@ -335,6 +335,51 @@ pub(crate) static FAILED_SEND_ENCRYPTED_LATENCY: Lazy<HistogramVec> = Lazy::new(
     .unwrap()
 });
 
+pub(crate) static NUM_TOTAL_HISTORICAL_SIGNATURE_GENERATORS: Lazy<IntGaugeVec> = Lazy::new(|| {
```
volovyks, line 338: Should we rename this to `SIGNATURE_GENERATOR`, `NEW_SIGNATURE_GENERATOR`, or something similar?
We should not expose how it is represented on Grafana in code. It can be reused, it can be per hour, not historical. It can be combined with other data and not be `TOTAL`.

I would rename all new and existing metrics so they represent what they are, and not how they are shown on in Grafana.

PR #808, opened 2024-08-08
chore: only use the GCP format within GCP

ChaoticTempest:
Moves the formatting of GCP related logging to be in the GCP conditional


volovyks:
What about his one?
https://github.com/near/mpc/pull/805

PR #809, opened 2024-08-08
Contract cleanup

volovyks:


chain-signatures/contract/src/lib.rs
```diff
@@ -612,7 +603,7 @@ impl VersionedMpcContract {
 
     // This function can be used to transfer the MPC network to a new contract.
     #[private]
-    #[init(ignore_state)]
```
volovyks, line 615: This function is less important since we plan to stick to 1 contract.
If we decide to move to another one - it should not have any state.

chain-signatures/contract/src/lib.rs
```diff
@@ -798,22 +790,6 @@ impl VersionedMpcContract {
         }
     }
 
-    #[private]
-    #[init(ignore_state)]
-    pub fn clean(keys: Vec<near_sdk::json_types::Base64VecU8>) -> Self {
```
volovyks, line 803: Not needed since we will not have any FAKs.

ChaoticTempest (approved):

PR #832, opened 2024-09-05
chore: reduce integration test boilerplate

ChaoticTempest:
This just reduces down some duplicated code and simplifies some things like account handling


volovyks (approved):
Looks good

PR #834, opened 2024-09-10
gas test

volovyks:


ailisp:
Nice finding! I would also test using a maximum sized request

volovyks:
@ailisp what do you mean by maximum sized request? Aren't they all the same size?

ppca (approved):

ailisp:
@volovyks i mean path. there was some clear_state_on_finish invoked by sign_helper spending around 6T gas while it usually spend 3T gas. Given the clear_state_on_finish is very simple, it maybe related to size of request

volovyks:
@ailisp good point, I've checked, and `path` size does [affect](https://testnet.nearblocks.io/txns/AM7B5Ym4M3Q5uEWN5LKS2ZL91h6f9fjxfXexUSguZ1MG) gas, but it is not significant for now.

PR #845, opened 2024-09-23
merge develop at rc4 into main branch

ppca:


volovyks:
Technically, it is not develope. Commit history can be messed up.

PR #847, opened 2024-09-23
Fixing prmetheus metrics definition

anshal-savla:


chain-signatures/node/src/metrics.rs
```diff
@@ -14,26 +14,26 @@ pub(crate) static NODE_RUNNING: Lazy<IntGaugeVec> = Lazy::new(|| {
     .unwrap()
 });
 
-pub(crate) static NUM_SIGN_REQUESTS: Lazy<IntGaugeVec> = Lazy::new(|| {
-    try_create_int_gauge_vec(
-        "multichain_sign_requests_count",
+pub(crate) static NUM_SIGN_REQUESTS_MINE: Lazy<CounterVec> = Lazy::new(|| {
```
volovyks, line 17: we should not rename this

chain-signatures/node/src/metrics.rs
```diff
@@ -14,26 +14,26 @@ pub(crate) static NODE_RUNNING: Lazy<IntGaugeVec> = Lazy::new(|| {
     .unwrap()
 });
 
-pub(crate) static NUM_SIGN_REQUESTS: Lazy<IntGaugeVec> = Lazy::new(|| {
-    try_create_int_gauge_vec(
+pub(crate) static NUM_SIGN_REQUESTS: Lazy<CounterVec> = Lazy::new(|| {
```
ppca, line 17: What's the difference between CounterVec and IntGaugeVec?

volovyks: You can not go down with counter. Also, there are some inconsistencies with the data. We are trying to find the reason.

volovyks (dismissed):

volovyks (approved):

PR #866, opened 2024-09-27
feat: add jwt_signature_pk_urls to state

esaminu:
This PR:

- Replaces `jwt_signature_pk_url` from `SignNodeState` and `LeadeState` with `jwt_signature_pk_urls` as comma separated list
- Modifies `get_public_keys` to fetch and handle public key responses with both [firebase format](https://www.googleapis.com/robot/v1/metadata/x509/securetoken@system.gserviceaccount.com) and standard jwks format [e.g. google](https://www.googleapis.com/oauth2/v3/certs)

The purpose is to add support to mpc for non firebase issuers e.g. google, apple etc. When a new issuer is added we need to ensure their public keys are also being fetched by appending their jwks url to this new env

### DIFF
--- a/mpc-recovery/src/oauth.rs
+++ b/mpc-recovery/src/oauth.rs
@@ -1,6 +1,8 @@
+use anyhow::{Context, Result};
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
 use jsonwebtoken::{Algorithm, DecodingKey};
 use serde::{Deserialize, Serialize};
-use std::collections::HashMap;
+use serde_json::Value;
 
 use crate::firewall::allowed::OidcProviderList;
 use crate::primitives::InternalAccountId;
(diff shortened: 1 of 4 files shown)


mpc-recovery/src/oauth.rs
```diff
@@ -99,13 +101,69 @@ impl IdTokenClaims {
...
-pub async fn get_pagoda_firebase_public_keys(
+pub async fn get_public_keys(
     client: &reqwest::Client,
-    jwt_signature_pk_url: &str,
-) -> anyhow::Result<Vec<String>> {
-    let response = client.get(jwt_signature_pk_url).send().await?;
-    let json: HashMap<String, String> = response.json().await?;
-    Ok(json.into_values().collect())
+    jwt_signature_pk_urls: &[String],
+) -> Result<Vec<String>> {
```
volovyks, line 107: It is better to create a map of keys: "provider" -> "[keys]".
Otherwise, somebody will create a token using `provider1`, but verify it with key 2.

volovyks (commented):
Looks good overall. I have concerns about security in the process of token verification.
Let's sync and discuss this.

mpc-recovery/src/sign_node/oidc.rs
```diff
@@ -98,6 +98,21 @@ impl OidcToken {
 
         Ok((header, claims, signature.into()))
     }
+
+    // NOTE: code taken directly from our implementation of token.decode but without the verification step
+    pub fn decode_unverified(&self) -> anyhow::Result<(jwt::Header, IdTokenClaims, String)> {
```
volovyks, line 103: Let's reuse it in decode

mpc-recovery/src/oauth.rs
```diff
@@ -99,13 +108,49 @@ impl IdTokenClaims {
...
+        Value::Object(obj) if obj.contains_key("keys") => parse_jwks_format(&obj),
+        Value::Object(obj) => parse_firebase_format(&obj),
+        _ => {
+            tracing::warn!("Unexpected response format from {}", jwks_url);
+            Ok(vec![])
+        }
+    }
+}
+
+fn parse_jwks_format(obj: &serde_json::Map<String, Value>) -> Result<Vec<String>> {
```
volovyks, line 130: Let's test this

volovyks (commented):
Looks good!

PR #869, opened 2024-10-02
Store Pre-signatures and Triples using Redis, refactor managers

### LINKED ISSUE #854: Persist pre-signatures
### Description
Currently, pre-signatures are stored in memory, which causes their loss on reboot.

volovyks:
This is the setup of Docker in the same container. Some people consider it a bad practice, but it may be a simpler solution for node validators. I have no strong opinion here.

It is expected to be executed with
```
docker run -d \
  --name my-redis-rust-app \
  -v /path/on/host/redis-data:/var/lib/redis \
  -p 6379:6379 \
  -p 8080:8080 \
  my-redis-rust-app
```

It will persist the apend file on the VM.

cc @kmaus-near @ppca


chain-signatures/node/redis.conf
```diff
@@ -0,0 +1,8 @@
+# Enable Append Only File (AOF) persistence
+appendonly yes
+
+# Set the appendfsync policy to flush the AOF every second
+appendfsync everysec
```
volovyks, line 5: We can afford to persist every call, but that will increase latency 4x. 1s should be a good compromise, we can lose a few presignatures now.

ppca:
@kmaus-near How does your draft redis setup work? Is it different or similar to this?

kmaus-near:
> @kmaus-near How does your draft redis setup work? Is it different or similar to this?

My homelab setup uses a dedicated VM running a Redis container vs. baking it into the multichain image, but other than that similar configuration. @volovyks and myself discussed this a bit yesterday, it's not exactly the "best practice" but if we are trying to simplify this for our partners, this may be a better route. Plus if we are to run it locally, we are then going to be running 3 containers on a single VM if we do not bake it into the image itself, which is not ideal resource wise or best practice wise. In short, both are not ideal, but for different reasons 😂.

I would suggest that we try to keep the code interacting with Redis as decoupled as possible, so if in the future we need to separate Redis from the base image, we can do so easily.

volovyks:
@kmaus-near all we need in code is the Redis URL, so it is not coupled. If we change how we manage Redis storage, most of the complexity will be on the Terraform side.

Dockerfile.multichain
```diff
@@ -19,9 +19,19 @@ RUN sed -i 's#target-dir = "../target"#target-dir = "target"#' .cargo/config.tom
...
 
 FROM debian:stable-slim as runtime
-RUN apt-get update && apt-get install --assume-yes libssl-dev ca-certificates curl
+RUN apt-get update && apt-get install --assume-yes libssl-dev ca-certificates curl redis-server
+
 RUN update-ca-certificates
 COPY --from=builder /usr/src/app/target/release/mpc-node /usr/local/bin/mpc-node
+
+# Create a script to start both Redis and the Rust app
+RUN echo "#!/bin/bash\nservice redis-server start\nexec mpc-node" > /start.sh \
```
kmaus-near, line 28: I think we may need to point to the location of the config file in order for it to not start in default config mode.

https://stackoverflow.com/a/23498203

Andy-commits-lgtm (approved):

ppca (commented):
presignature storage logic mostly look good to me.

.github/workflows/multichain-integration.yml
```diff
@@ -39,11 +39,12 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Pull Relayer & Sandbox Docker Images
+      - name: Pull Relayer, Sandbox, Redis Docker Images
         run: |
           docker pull ghcr.io/near/os-relayer:12ba6e35690df3979fce0b36a41d0ca0db9c0ab4
```
ChaoticTempest: wait, why do we pull relayer in our multichain integration?

Dockerfile.multichain
```diff
@@ -19,9 +19,19 @@ RUN sed -i 's#target-dir = "../target"#target-dir = "target"#' .cargo/config.tom
...
+# Create a script to start both Redis and the Rust app
+RUN echo "#!/bin/bash\nservice redis-server start ./redis.conf\nexec mpc-node" > /start.sh \
+    && chmod +x /start.sh
+
 WORKDIR /usr/local/bin
 
-ENTRYPOINT [ "mpc-node" ]
\ No newline at end of file
+# Expose Redis port (6379) and mpc node's port
+EXPOSE 6379 8080
```
ChaoticTempest: hmm, how come we didn't need to expose the node's port before

chain-signatures/node/src/protocol/message.rs
```diff
@@ -309,7 +309,7 @@ impl MessageHandler for RunningState {
             }
 
             let protocol = match presignature_manager
-                .get_or_generate(
+                .get_or_start_protocol(
```
ChaoticTempest: hmm, generate is basically start protocol though. I would rather we keep it as what we had before, because we also have the `generate` function

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -23,12 +27,94 @@ use near_account_id::AccountId;
 pub type PresignatureId = u64;
 
 /// A completed presignature.
+#[derive(Clone)]
 pub struct Presignature {
```
ChaoticTempest: `Presignature` should never be clone'able since that introduces reusage of presignature. Why do we even need it at all?

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -23,12 +27,94 @@ use near_account_id::AccountId;
...
+            id: 1,
+            output: PresignOutput {
+                big_r: <Secp256k1 as CurveArithmetic>::AffinePoint::default(),
+                k: <Secp256k1 as CurveArithmetic>::Scalar::ZERO,
+                sigma: <Secp256k1 as CurveArithmetic>::Scalar::ONE,
+            },
+            participants: vec![Participant::from(1), Participant::from(2)],
+        }
+    }
+}
```
ChaoticTempest: we should not be having defaults for presignatures either. In what case is it used? If for testing, we should have a dummy function instead of default

infra/multichain-dev/main.tf
```diff
@@ -14,6 +14,14 @@ module "gce-container" {
     args  = ["start"]
     port  = "3000"
 
+    volumeMounts = [
```
ChaoticTempest, line 16: don't we need to do this for more than just `multichain-dev/`?

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -23,12 +27,94 @@ use near_account_id::AccountId;
...
+            serde_json::to_value(self.output.big_r).map_err(serde::ser::Error::custom)?,
+        );
+        output_map.insert(
+            "k".to_string(),
+            serde_json::to_value(self.output.k).map_err(serde::ser::Error::custom)?,
+        );
+        output_map.insert(
+            "sigma".to_string(),
+            serde_json::to_value(self.output.sigma).map_err(serde::ser::Error::custom)?,
+        );
```
ChaoticTempest: we shouldn't be encoding this with serde_json as that has overhead. Is there a specific reason why we don't just serialize each field of output?
```
        let mut state = serializer.serialize_struct("Presignature", 5)?;
        state.serialize_field("id", &self.id)?;
        state.serialize_field("output_big_r", &self.output.big_r)?;
        state.serialize_field("output_k", &self.output.k)?;
        state.serialize_field("output_sigma", &self.output.sigma)?;
        state.serialize_field("participants", &self.participants)?;
        state.end()
```

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -530,6 +696,14 @@ impl PresignatureManager {
             }
         });
 
+        for presignature in new_presignatures {
+            self.insert(presignature).await;
+        }
+
+        for presignature in new_mine_presignatures {
+            self.insert_mine(presignature).await;
+        }
```
ChaoticTempest, line 649: not important for now, but would it better if we could insert all of them into redis at once instead of one by one?

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -0,0 +1,134 @@
...
+pub trait PresignatureStorage {
+    fn insert(&mut self, presignature: Presignature) -> PresigResult<()>;
+    fn insert_mine(&mut self, presignature: Presignature) -> PresigResult<()>;
+    fn contains(&mut self, id: &PresignatureId) -> PresigResult<bool>;
+    fn contains_mine(&mut self, id: &PresignatureId) -> PresigResult<bool>;
+    fn take(&mut self, id: &PresignatureId) -> PresigResult<Option<Presignature>>;
+    fn take_mine(&mut self) -> PresigResult<Option<Presignature>>;
+    fn count_all(&mut self) -> PresigResult<usize>;
+    fn count_mine(&mut self) -> PresigResult<usize>;
+}
```
ChaoticTempest: we don't need a trait at all because we use soley use Redis and have no in-memory alternative

integration-tests/chain-signatures/src/lib.rs
```diff
@@ -232,13 +232,19 @@ pub async fn setup(docker_client: &DockerClient) -> anyhow::Result<Context<'_>>
     let datastore =
         crate::containers::Datastore::run(docker_client, docker_network, gcp_project_id).await?;
 
+    let redis_container = crate::containers::Redis::run(docker_client, docker_network).await?;
+    let redis_url = redis_container.local_address.clone();
+
+    tracing::info!(redis_url = %redis_url, "started Redis container");
```
ChaoticTempest:
```suggestion
    tracing::info!(%redis_url, "started Redis container");
```

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -135,25 +225,131 @@ impl PresignatureManager {
...
+            .map_err(|e| {
+                tracing::error!(?e, "failed to count all presignatures");
+            })
+            .unwrap_or(0)
     }
 
     /// Returns the number of unspent presignatures assigned to this node.
-    pub fn my_len(&self) -> usize {
-        self.mine.len()
+    pub async fn count_mine(&self) -> usize {
```
ChaoticTempest: why change it all to `count_*`? Now it's gonna be inconsistent with what we have in triple manager

infra/multichain-dev/main.tf
volovyks: cc @kmaus-near

chain-signatures/node/src/protocol/presignature.rs
volovyks: It is possible in Redis. I may add it later with atomicity PR (ability to read, write, and modify Redis state in a single atomic transaction).

chain-signatures/node/src/protocol/presignature.rs
volovyks: done

.github/workflows/multichain-integration.yml
volovyks: because we can
removed

chain-signatures/node/src/protocol/message.rs
volovyks: Renamed to `get_or_start_generation`.
When you call `presignature_manager.get_or_generate` you are expecting a presignature, not a protocol.

infra/multichain-dev/main.tf
kmaus-near: Yes, when it's tested and ready to move into main branch. Eventually I would like to automate our own TF deployments for production, so this is more of a development workflow question we should eventually talk about.

chain-signatures/node/src/storage/presignature_storage.rs
volovyks: Ok
And we do need Arc now

chain-signatures/node/src/protocol/presignature.rs
volovyks: Yes, for testing
ok, security first

chain-signatures/node/src/protocol/presignature.rs
volovyks: yeah, for testing, ok security first

chain-signatures/node/src/protocol/presignature.rs
volovyks: agree, simlified

Dockerfile.multichain
kmaus-near: Also wondering why 8080 is now exposed. Port is set at 3000 for the mpc-node service by terraform. we will need to expose 6379 though for redis.

Dockerfile.multichain
```diff
@@ -19,9 +19,22 @@ RUN sed -i 's#target-dir = "../target"#target-dir = "target"#' .cargo/config.tom
 RUN cargo build --release --package mpc-node
 
 FROM debian:stable-slim as runtime
-RUN apt-get update && apt-get install --assume-yes libssl-dev ca-certificates curl
-RUN update-ca-certificates
+RUN apt-get update && apt-get install --assume-yes libssl-dev ca-certificates curl redis-server
+
+RUN update-ca-certificates \
+    chown redis:redis /data
```
kmaus-near: I needed this /data directory to be owned by redis, so it can store data.

chain-signatures/node/redis.conf
kmaus-near, line 1: I needed to add the rest of the config. I have changed the AOF parameters as well as using the /data directory inside the container for persistence.

infra/multichain-dev/main.tf
```diff
@@ -58,9 +65,22 @@ module "gce-container" {
...
+        value = var.redis_url
       }
     ])
   }
+
+  volumes = [
+      {
+        name = "host-path"
+        hostPath = {
+          path = "/var/redis"
```
kmaus-near, line 80: had to change this from /etc to /var. In container optimized OS, /etc is a stateless directory and ships with a default configuration. /var is persistent and gets mounted to /mnt/stateful-partition

chain-signatures/node/src/protocol/message.rs
ChaoticTempest: hmm, that's not right either though. `get_or_generate` never returned us the presignature, it always returned us the protocol

volovyks: yes, it never did, but it is expected from the `get_or_generate` name semantics, which is wrong

volovyks:
I've fixed an issue with Redis integration tests. Most of the tests fail because of the cashing (probably) issue.
TF setup was tested on one of the dev nodes (thanks @kmaus-near ).

volovyks:
Merged triple storage into this PR:

What was done:
- Removed duplicate storage (DB and memory) approach
- Datastore replaced with Redis
- Refactoring of TripleManager

Other:
- I removed StuckMonitor because it does not fit well into the design with Redis. I assume it was used for testing. I can add a lightweight version in Triple and Presignature managers if requested.

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -407,20 +407,23 @@ impl CryptographicProtocol for RunningState {
...
             .with_label_values(&[my_account_id.as_str()])
-            .set(presignature_manager.len() as i64);
+            .set(presignature_manager.count_all().await as i64);
         crate::metrics::NUM_PRESIGNATURE_GENERATORS_TOTAL
             .with_label_values(&[my_account_id.as_str()])
-            .set(presignature_manager.potential_len() as i64 - presignature_manager.len() as i64);
+            .set(
+                presignature_manager.count_potential().await as i64
+                    - presignature_manager.count_all().await as i64,
+            );
```
ChaoticTempest: ok, this naming is now pretty confusing. When I see `count_all` I would assume it includes potential as well. This calculation tells me it does not, but then `all` is not inclusive and is confusing.

Not a fan of the word `count` word because `len` is so much more standard in rust. I now don't see a reason to change what we had before.

I say we have this consistent naming instead:
- `len`: number of triple/presignatures
- `len_mine`: number of mine triple/presignatures
- `len_generators`: number of generators
- `len_potential`: current number of triple/presignature + generators

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -135,25 +185,133 @@ impl PresignatureManager {
         }
     }
 
+    pub async fn insert(&mut self, presignature: Presignature) {
+        tracing::info!(id = ?presignature.id, "inserting presignature");
```
ChaoticTempest: I'd say making this debug. Gonna flood our logs this way

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -135,25 +185,133 @@ impl PresignatureManager {
...
+        // Remove from taken list if it was there
+        self.gc.remove(&presignature.id);
+        if let Err(e) = self.presignature_storage.write().await.insert(presignature) {
+            tracing::error!(?e, "failed to insert presignature");
+        }
+    }
+
+    pub async fn insert_mine(&mut self, presignature: Presignature) {
+        tracing::info!(id = ?presignature.id, "inserting mine presignature");
```
ChaoticTempest: same with this one

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -403,48 +565,15 @@ impl PresignatureManager {
...
     /// Pokes all of the ongoing generation protocols and returns a vector of
     /// messages to be sent to the respective participant.
     ///
     /// An empty vector means we cannot progress until we receive a new message.
-    pub fn poke(&mut self) -> Vec<(Participant, PresignatureMessage)> {
+    pub async fn poke(&mut self) -> Vec<(Participant, PresignatureMessage)> {
         let mut messages = Vec::new();
         let mut errors = Vec::new();
+        let mut new_presignatures = Vec::new();
+        let mut new_mine_presignatures = Vec::new();
```
ChaoticTempest, line 558: now that I see this, it's better for us to just have an extra bool field denoting the ownership of the presignature even though we already maintain a list of mine presignatures in redis. Would simplify the code a bit

volovyks: Redis is not designed for such data structures. For example, it will need to deserialize all triples to count mine.

chain-signatures/node/src/protocol/cryptography.rs
volovyks: I agree that `len()` is a good name when we work directly with a map. But here it is usually called as
```
triple_manager.count_mine()
```
In this conext `len()` is a bit ambiguous.

Agree about `count_all()`, will rename it to `count()` or maybe `count_generated()` to differentiate between generation protocols and pre-signatures that are ready to use.

volovyks: Hm, if we add `_generated` it make sense with `len` too.

ChaoticTempest:
https://github.com/near/mpc/pull/914 should fix the issues you were seeing @volovyks. Too many write locks were being acquired at the same time

ppca:
LGTM!

ppca (dismissed):

ppca:
@ChaoticTempest was the reason behind removing Arc and RwLock for triple/presignature storage because redis itself can already handle the concurrent read and write with its own locking mechanism?

kmaus-near (approved):

ChaoticTempest:
@ppca yup, when sending items to be stored to redis, it's going to be over the wire anyways, so mutability does not matter at all here, thus we don't need `RwLock` at all. The redis pool struct actually has an internal `Arc`, so no need for us to manage it ourself

PR #883, opened 2024-10-10
add timeout to requests to nodes and add metric for iter cnts

ppca:


chain-signatures/node/src/metrics.rs
```diff
@@ -398,6 +398,15 @@ pub(crate) static SIGNATURE_PUBLISH_RESPONSE_ERRORS: Lazy<CounterVec> = Lazy::ne
     .unwrap()
 });
 
+pub(crate) static PROTOCOL_LATENCY_ITER_CNT: Lazy<CounterVec> = Lazy::new(|| {
+    try_create_counter_vec(
```
volovyks, line 402: Do you want to measure latency or count iterations? Let's rename this accordingly and use counter if it's the later one.

volovyks (commented):
Good finding!
What is the expected behaviour now? From what I see, instead of being stuck, we will start having relatively big delayes. Especially in /msg endpoint. Can we aford that? One unresponsive node can slow things down.

ChaoticTempest (dismissed):

ppca:
> Good finding! What is the expected behaviour now? From what I see, instead of being stuck, we will start having relatively big delayes. Especially in /msg endpoint. Can we aford that? One unresponsive node can slow things down.

That is a timeout, if the request finish before 5s, it will return much sooner than 5s. So it unstuck when /msg is not available, and it will be as fast if /msg is working properly.

volovyks (approved):

kmaus-near (approved):

PR #887, opened 2024-10-10
modify is_stable to be indexer running and latest block timestamp not behind

### LINKED ISSUE #871: is_stable is true for nodes with lagging block heights
If a node starts at a very early block height and has been syncing but is still way behind, is_stable will still show true

ppca:
indexer progressing = local indexer block height last update timestamp is within threshold
indexer caught up = my block height >= latest height from near rpc endpoint - 50

This will fix the case where some nodes are catching up but still not update to date, and they got involved in the signature generation

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -59,7 +59,7 @@ pub struct Options {
     #[clap(
         long,
         env("MPC_INDEXER_BLOCK_HEIGHT_LAG_THRESHOLD"),
-        default_value = "50"
+        default_value = "200"
     )]
     pub block_height_lag_threshold: u64,
 }
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -1,6 +1,7 @@
 use crate::gcp::error::DatastoreStorageError;
 use crate::gcp::GcpService;
 use crate::protocol::{SignQueue, SignRequest};
+use crate::rpc_client;
 use crate::types::LatestBlockHeight;
 use crypto_shared::{derive_epsilon, ScalarExt};
 use k256::Scalar;
@@ -53,6 +54,14 @@ pub struct Options {
     /// The threshold in seconds to check if the indexer needs to be restarted due to it stalling.
     #[clap(long, env("MPC_INDEXER_RUNNING_THRESHOLD"), default_value = "300")]
     pub running_threshold: u64,
+
+    /// The threshold in block height lag to check if the indexer has caught up.
+    #[clap(
+        long,
+        env("MPC_INDEXER_BLOCK_HEIGHT_LAG_THRESHOLD"),
+        default_value = "50"
+    )]
+    pub block_height_lag_threshold: u64,
 }
 
 impl Options {
@@ -68,6 +77,8 @@ impl Options {
             self.behind_threshold.to_string(),
             "--running-threshold".to_string(),
             self.running_threshold.to_string(),
+            "--block-height-lag-threshold".to_string(),
+            self.block_height_lag_threshold.to_string(),
         ];
 
         if let Some(s3_url) = self.s3_url {
@@ -105,10 +116,16 @@ pub struct Indexer {
     last_updated_timestamp: Arc<RwLock<Instant>>,
     running_threshold: Duration,
     behind_threshold: Duration,
+    block_height_lag_threshold: u64,
+    rpc_client: near_fetch::Client,
 }
 
 impl Indexer {
-    fn new(latest_block_height: LatestBlockHeight, options: &Options) -> Self {
+    fn new(
+        latest_block_height: LatestBlockHeight,
+        options: &Options,
+        rpc_client: near_fetch::Client,
+    ) -> Self {
         tracing::info!(
             "creating new indexer, latest block height: {}",
             latest_block_height.block_height
@@ -118,6 +135,8 @@ impl Indexer {
             last_updated_timestamp: Arc::new(RwLock::new(Instant::now())),
             running_threshold: Duration::from_secs(options.running_threshold),
             behind_threshold: Duration::from_secs(options.behind_threshold),
+            block_height_lag_threshold: options.block_height_lag_threshold,
+            rpc_client,
         }
     }
 
@@ -126,7 +145,7 @@ impl Indexer {
         self.latest_block_height.read().await.block_height
     }
 
-    /// Check whether the indexer is on track with the latest block height from the chain.
+    /// Check whether the indexer block height has been updated recently.
     pub async fn is_on_track(&self) -> bool {
         self.last_updated_timestamp.read().await.elapsed() <= self.behind_threshold
     }
@@ -138,7 +157,17 @@ impl Indexer {
 
     /// Check whether the indexer is behind with the latest block height from the chain.
     pub async fn is_behind(&self) -> bool {
-        self.last_updated_timestamp.read().await.elapsed() > self.behind_threshold
+        let network_latest_height = rpc_client::fetch_latest_block_height(&self.rpc_client).await;
+        if let Ok(network_latest_height) = network_latest_height {
+            self.latest_block_height().await
+                < network_latest_height - self.block_height_lag_threshold
+        } else {
+            false
+        }
+    }
+
+    pub async fn is_stable(&self) -> bool {
+        !self.is_behind().await && self.is_on_track().await
     }
 
     async fn update_block_height(
@@ -287,6 +316,7 @@ pub fn run(
     node_account_id: &AccountId,
     queue: &Arc<RwLock<SignQueue>>,
     gcp_service: &crate::gcp::GcpService,
+    rpc_client: near_fetch::Client,
     rt: &tokio::runtime::Runtime,
 ) -> anyhow::Result<(JoinHandle<anyhow::Result<()>>, Indexer)> {
     tracing::info!(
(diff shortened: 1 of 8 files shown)


chain-signatures/node/src/mesh/mod.rs
```diff
@@ -53,7 +53,15 @@ impl Mesh {
...
         for (participant, info) in self.active_participants().iter() {
-            if self.connections.is_participant_stable(participant).await {
+            if self
+                .connections
+                .is_participant_indexer_progressing(participant)
+                .await
+                && self
+                    .connections
+                    .is_participant_indexer_caught_up(participant)
+                    .await
```
ChaoticTempest, line 63: we should just encapsulate all this logic in `is_participant_stable` function anyways

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -161,14 +162,61 @@ impl Pool {
...
+            .and_then(|state| match state {
+                StateView::Running {
+                    latest_block_height,
+                    ..
+                } => Some(*latest_block_height),
+                _ => None,
+            })
+            .unwrap_or(0);
+
+        (max_block_height - my_block_height) < 50
```
ChaoticTempest, line 220: the range should be configurable

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -161,14 +162,61 @@ impl Pool {
...
+                    ..
+                } = state
+                {
+                    Some(*latest_block_height)
+                } else {
+                    None
+                }
+            })
+            .max()
+            .unwrap_or(0)
```
ChaoticTempest, line 182: this should just return an Option or Result instead of unwrapping into zero since we end up checking for zero later

ChaoticTempest (commented):
Looks good for the most part, but we really should not be changing the field names for messages

volovyks (commented):
Phuong's comments are spot on. Overall, it looks good. However, we should be mindful that a single node could potentially disrupt the system by reporting a block height that is higher than the actual value.

ppca:
@ChaoticTempest @volovyks I changed the logic:
1) get max block height from near rpc instead of from participants, this means as long as the node is honest, it will be compared against the true latest block height;
2) block_height_lag_threshold will be one field in indexer and is configurable thru environment variable;
3) instead of changing the definition of is_participant_stable(), I added a function is_stable() in indexer, that checks a) block heights has been updated lately, b) I am not more than  block_height_lag_threshold behind the max block height form near rpc;
4) given 3), the is_stable field in StateView is directly updated to be `is_stable = self.indexer.is_stable().await`, which will encapsulate both progressing and height being caught up

ppca:
I do want to add a timeout to all the near rpc calls we have in our code, is there an easy way to do it for near_fetch::Client? @ChaoticTempest

chain-signatures/node/src/indexer.rs
```diff
@@ -138,7 +157,17 @@ impl Indexer {
 
     /// Check whether the indexer is behind with the latest block height from the chain.
     pub async fn is_behind(&self) -> bool {
-        self.last_updated_timestamp.read().await.elapsed() > self.behind_threshold
+        let network_latest_height = rpc_client::fetch_latest_block_height(&self.rpc_client).await;
+        if let Ok(network_latest_height) = network_latest_height {
+            self.latest_block_height().await
+                < network_latest_height - self.block_height_lag_threshold
+        } else {
+            false
```
volovyks, line 165: Should we return `true` if we fail to get the latest block height?

chain-signatures/node/src/indexer.rs
```diff
@@ -53,6 +54,14 @@ pub struct Options {
     /// The threshold in seconds to check if the indexer needs to be restarted due to it stalling.
     #[clap(long, env("MPC_INDEXER_RUNNING_THRESHOLD"), default_value = "300")]
     pub running_threshold: u64,
+
+    /// The threshold in block height lag to check if the indexer has caught up.
```
volovyks, line 58: Let's discuss a strategy for configuration at our next meeting. Do we want to put everything in the contract?

ChaoticTempest:
> I do want to add a timeout to all the near rpc calls we have in our code, is there an easy way to do it for near_fetch::Client? @ChaoticTempest

You have to use either `transact_async` and do the check manually, or keep `transact` and use `tokio::time::timeout`

chain-signatures/node/src/indexer.rs
ChaoticTempest: yeah should be `true` if it's behind

chain-signatures/node/src/indexer.rs
ChaoticTempest: yeah, we should move all these indexer configurations into the contract to make it easier to configure

chain-signatures/node/src/indexer.rs
```diff
@@ -53,6 +54,14 @@ pub struct Options {
...
+    #[clap(
+        long,
+        env("MPC_INDEXER_BLOCK_HEIGHT_LAG_THRESHOLD"),
+        default_value = "50"
```
ChaoticTempest, line 62: hmm, not sure if 50 is the best. Might be too little to say it's behind. Have you tested longer lag threshold like 100 or 500?

volovyks: I was thinking about a lower number.
50 blocks is ~50 seconds. Are we often hitting that threshold?

If the node is 50 seconds behind, it fails all the assigned requests. Yes, it can still participate in other node protocols, but I would not consider it as "stable".

ppca: I actually just looked at our dev where when it's generating signatures alright, the heights are within 50 from one another.

chain-signatures/node/src/indexer.rs
ppca: Yeah I was thinking the same when adding timeout options today.

chain-signatures/node/src/indexer.rs
ppca: I thought about it again. I think we should use a bigger value than 50. This lag is comparing with the latest block fetched from near rpc, and if lake has any delays, all nodes will be identified as unstable. 50s is probably too strict.
I'm thinking 200, as it is a bit smaller than the longest a signature request can wait in yield/resume, which means if nodes are <200 blocks behind, they'll still be able to answer the request in time.

ppca:
When doing this PR, I actually realized we only need to proposer to be stable. This is because 1) proposer is the one starting the signature protocol, other nodes just joining and they don't check if they have that request locally; 2) proposer is deterministic for each sign request, so there's no risk of unstable nodes later re-starting protocol for the same signature request, because it will only be that same proposer always. 
So by limiting lag from rpc latest block to 200, we make sure proposers will see the sign request within 3.5 mins, and have another 1 min to respond().
If that lag is too big, that node will not make it in time anyways.
[100-200] might all be good.
@ChaoticTempest @volovyks

ppca:
I also realized with our current implementation, when a node catches up on heights, it is likely to start protocols to generate signature for an old sign request, because the stable participants set change, and if you look at logic here: https://github.com/near/mpc/blob/6557f017246d467fd5f9cee4ee3061394d19140a/chain-signatures/node/src/protocol/signature.rs#L118
The subset and proposer for the signature can be different from last time and thus this node who just caught up could end up being the proposer and respond() again.
Of course the range of sign requests will be limited to whatever lag threshold we allow in this PR.

chain-signatures/node/src/indexer.rs
ChaoticTempest: yup, lake can be delayed a fair bit, so yeah let's do something like 200. Lines up very well with yield/resume

ChaoticTempest:
wait, we don't need to keep track of the block height from rpc vs our current block height from indexer, we can just use the block timestamp to check how far behind we are in comparison with our current time. So we don't need to add this fetching of the block from RPC which can also be delayed by a couple seconds

> When doing this PR, I actually realized we only need to proposer to be stable. This is because 1) proposer is the one starting the signature protocol, other nodes just joining and they don't check if they have that request locally; 2) proposer is deterministic for each sign request, so there's no risk of unstable nodes later re-starting protocol for the same signature request, because it will only be that same proposer always. So by limiting lag from rpc latest block to 200, we make sure proposers will see the sign request within 3.5 mins, and have another 1 min to respond(). If that lag is too big, that node will not make it in time anyways. [100-200] might all be good.

Even less strict -- it doesn't have to be the proposer, it can be any stable participant because it doesn't matter who responds with the signature.


> I also realized with our current implementation, when a node catches up on heights, it is likely to start protocols to generate signature for an old sign request, because the stable participants set change, and if you look at logic here:
> 
> https://github.com/near/mpc/blob/6557f017246d467fd5f9cee4ee3061394d19140a/chain-signatures/node/src/protocol/signature.rs#L118
> 
> 
> The subset and proposer for the signature can be different from last time and thus this node who just caught up could end up being the proposer and respond() again.
> Of course the range of sign requests will be limited to whatever lag threshold we allow in this PR.

Yeah, we can just reject a signature request ourselves based on our threshold timing with the block timestamp

ppca:
> we can just use the block timestamp to check how far behind we are in comparison with our current time

That's cool! How can I do that? @ChaoticTempest

ChaoticTempest:
@ppca in `handle_block`, it should be `block.header().timestamp_nanosec()`

volovyks (approved):

ChaoticTempest (approved):

PR #889, opened 2024-10-13
use tokio::time::timeout to timeout fetch state and send encrypted

### LINKED ISSUE #873: Fix 12 node system
### Description
It may be slow, but it should be operational with high uptime.

ppca:
The previous implementation caused troubles: 1) 1s timeout have already stuck the protocol; 2) 500ms timeout still see many timeouts on /state endpoint.
I am guessing the .timeout() that came with reqwest package cannot successfully time out an async call, the counting of time may not have considered task/thread switching.
So I switched to using tokio::time::timeout and now problem solved. 
Dev success increased from 70% to 100%.

This change sets the timeout to 1000ms for both fetch /state and for send encrypted. Both can be tuned as an environment variable.

### DIFF
--- a/chain-signatures/node/src/cli.rs
+++ b/chain-signatures/node/src/cli.rs
@@ -2,7 +2,7 @@ use crate::config::{Config, LocalConfig, NetworkConfig, OverrideConfig};
 use crate::gcp::GcpService;
 use crate::protocol::{MpcSignProtocol, SignQueue};
 use crate::storage::triple_storage::LockTripleNodeStorageBox;
-use crate::{indexer, storage, web};
+use crate::{http_client, indexer, mesh, storage, web};
 use clap::Parser;
 use local_ip_address::local_ip;
 use near_account_id::AccountId;
@@ -63,6 +63,10 @@ pub enum Cli {
         /// referer header for mainnet whitelist
         #[arg(long, env("MPC_CLIENT_HEADER_REFERER"), default_value(None))]
         client_header_referer: Option<String>,
+        #[clap(flatten)]
+        mesh_options: mesh::Options,
+        #[clap(flatten)]
+        message_options: http_client::Options,
     },
 }
 
@@ -83,6 +87,8 @@ impl Cli {
                 storage_options,
                 override_config,
                 client_header_referer,
+                mesh_options,
+                message_options,
             } => {
                 let mut args = vec![
                     "start".to_string(),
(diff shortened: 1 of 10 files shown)


chain-signatures/node/src/mesh/connection.rs
```diff
@@ -49,32 +49,37 @@ impl Pool {
...
-                .send()
-                .await
-            else {
-                tracing::warn!(
-                    "Pool.ping resp err participant {:?} url {}",
-                    participant,
-                    url
-                );
-                continue;
+            let work = self.http.get(url.clone()).send();
```
volovyks, line 52: Since we have issues with node stability, should we add this info to metrics? Each node can report the number of active participants. We can even add an alert on this.

ppca: sounds good! I'll do it in another PR since this PR will already be big

ppca:
as discussed with @volovyks , since we might replace the client later with sockets etc, we will not refactor the timeouts into contract.

chain-signatures/node/src/cli.rs
```diff
@@ -63,6 +63,10 @@ pub enum Cli {
         /// referer header for mainnet whitelist
         #[arg(long, env("MPC_CLIENT_HEADER_REFERER"), default_value(None))]
         client_header_referer: Option<String>,
+        #[clap(flatten)]
+        mesh_options: mesh::Options,
+        #[clap(flatten)]
+        message_options: http_client::Options,
```
ChaoticTempest, line 69: hmm, both these should just be in the contract config, especially `message_options` since that is very similar to our generation timeout configuration

ChaoticTempest:
> as discussed with @volovyks , since we might replace the client later with sockets etc, we will not refactor the timeouts into contract.

I see, then let's not bake these sort of configuration into terraform at all and just keep with the defaults in the CLI itself for now

chain-signatures/node/src/cli.rs
```diff
@@ -63,6 +63,10 @@ pub enum Cli {
         /// referer header for mainnet whitelist
         #[arg(long, env("MPC_CLIENT_HEADER_REFERER"), default_value(None))]
         client_header_referer: Option<String>,
+        #[clap(flatten)]
+        mesh_options: mesh::Options,
```
volovyks, line 67: Do partners need to update their terraform setup?

volovyks (approved):
Looks good!

PR #890, opened 2024-10-14
Ethereum contract

ailisp:
First part of near/transfer#7 . The interface can be found in ChainSignatures.test.js. The precomputed value is taken from rust contract / test output in [`bo/eth-reference`](https://github.com/near/mpc/compare/bo/eth-reference?expand=1) branch

Refer to README in this PR to see how this part works. scripts mentioned in README is for prototype and a reference impl, they will be removed when actual impl in mpc node is finished


chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    function deriveKey(PublicKey memory _publicKey, uint256 epsilon) public pure returns (PublicKey memory) {
+        // G * epsilon + publicKey
+        (uint256 epsilonGx, uint256 epsilonGy) = Secp256k1.ecMul(epsilon, Secp256k1.GX, Secp256k1.GY);
+        (uint256 resultX, uint256 resultY) = Secp256k1.ecAdd(epsilonGx, epsilonGy, _publicKey.x, _publicKey.y);
+        return PublicKey(resultX, resultY);
+    }
+
+    function deriveEpsilon(string memory path, address requester) public pure returns (uint256) {
```
volovyks, line 71: This function should not be a part of the contract interface.
Let's move all public functions to the top with sign and pk functions on top. That would be convenient for people who want to integrate with this contract.

volovyks: There are a few more places where functions should not be public.

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -0,0 +1,160 @@
...
+        string memory requesterStr = Strings.toHexString(uint256(uint160(requester)), 20);
+        string memory epsilonString = string.concat("near-mpc-recovery v0.2.0 epsilon derivation:", requesterStr, ",", path);
+        console.log("Epsilon String:", epsilonString);
+        bytes32 epsilonBytes = keccak256(bytes(epsilonString));
+        uint256 epsilon = uint256(epsilonBytes);
+        return epsilon;
+    }
+
+    function sign(bytes32 payload, string memory path) external payable returns (bytes32) {
```
volovyks: It is important to have a key version parameter here, even if this logic has not been implemented yet to avoid breaking changes later.
Should we use structs in function APIs? I am not sure what the best practices in Solidity are.

chain-signatures/contract-eth/contracts/FastEcMul.sol
```diff
@@ -0,0 +1,526 @@
...
+import "./EllipticCurve.sol";
+
+/**
+ ** @title Fast Elliptic Curve Multiplication Library
+ ** @dev Library providing the following speed ups to an elliptic curve multiplication operation.
+ **      - wNAF scalar representation
+ **      - scalar decomposition through endomorphism
+ ** This library does not check whether the inserted points belong to the curve
+ ** `isOnCurve` function should be used by the library user to check the aforementioned statement.
+ ** @author Witnet Foundation
```
volovyks, line 14: Is this and other tools available as a library?

chain-signatures/contract-eth/ignition/modules/chainSignatures.js
```diff
@@ -0,0 +1,15 @@
+const { buildModule } = require("@nomicfoundation/hardhat-ignition/modules");
+
+const DEFAULT_PUBLIC_KEY = {
+  x: "0xfc115813e59a914d7566a4b1d4263048faf6b6dab6893c4d65d39fb123da5651",
```
volovyks, line 4: Is it something specific?

chain-signatures/contract-eth/test/ChainSignatures.test.js
```diff
@@ -0,0 +1,177 @@
...
+    const publicKeyArray = [4, 252, 17, 88, 19, 229, 154, 145, 77, 117, 102, 164, 177, 212, 38, 48, 72, 250, 246, 182, 218, 182, 137, 60, 77, 101, 211, 159, 177, 35, 218, 86, 81, 34, 58, 8, 114, 107, 154, 223, 0, 50, 161, 52, 118, 17, 227, 94, 159, 20, 183, 168, 231, 238, 49, 165, 217, 4, 25, 10, 78, 246, 250, 71, 225];    
+    const x = publicKeyArray.slice(1, 33);
+    const y = publicKeyArray.slice(33);
+    const xHex = '0x' + x.map(byte => byte.toString(16).padStart(2, '0')).join('');
+    const yHex = '0x' + y.map(byte => byte.toString(16).padStart(2, '0')).join('');
+    console.log("===", xHex, yHex);
+    ChainSignatures = await ethers.getContractFactory("ChainSignatures");
+    publicKey = { x: xHex, y: yHex };
+    chainSignatures = await ChainSignatures.deploy(publicKey);
+    // await chainSignatures.deployed();
```
volovyks: do we still need this await?

chain-signatures/contract-eth/scripts/sign.js
```diff
@@ -0,0 +1,81 @@
+const hre = require("hardhat");
+
+async function main() {
```
volovyks, line 3: Not sure what we should have in the end. JS tests may be hard to integrate into our current infra and harder to support.

chain-signatures/contract-eth/scripts/responder.js
```diff
@@ -0,0 +1,53 @@
+const hre = require("hardhat");
+
+async function main() {
```
volovyks, line 3: Do you want to keep this after you add a real indexer to the main system?

volovyks (dismissed):
Great work!

chain-signatures/contract-eth/contracts/ChainSignatures.sol
ailisp: Noted, they are otherwise not easy to test with, so i leave them public in the proof of concept. Will be refactored right after the proof of concept works

chain-signatures/contract-eth/contracts/ChainSignatures.sol
ailisp: Will add a key version. Struct can be used, costing a bit more gas.

chain-signatures/contract-eth/scripts/responder.js
ailisp: No, it is reference client before integration with node is done

chain-signatures/contract-eth/contracts/FastEcMul.sol
ailisp: This one is, the other secp256k1 one is not. We probably need our custom related optimization on this file, trim it, and gas optimize it further

chain-signatures/contract-eth/ignition/modules/chainSignatures.js
ailisp: no, for testing

chain-signatures/contract-eth/scripts/sign.js
ailisp: did more research, currently hardhat is still most common option in eth and most evm compatible chains. This file and responder.js is temporary one that will be dropped after integrate with mpc node

chain-signatures/contract-eth/test/ChainSignatures.test.js
ailisp: Removed

chain-signatures/contract-eth/contracts/ChainSignatures.sol
ailisp: added

ailisp:
@volovyks ptal. Addressed most of your points. Other not addressed points are valid too, will update them after finishing the indexer & gateway parts

chain-signatures/contract-eth/scripts/sign.js
volovyks: ok, we should use what is the industry standard

volovyks (approved):

PR #894, opened 2024-10-15
add /msg to criteria of active participants

ppca:
added `test_empty_msg()` in the logic of establishing active participants.
Now to qualify as an active participant:
1) /state returns response fine;
2) sending empty message thru send_encrypted() returns Ok(());

### DIFF
--- a/chain-signatures/node/src/mesh/connection.rs
+++ b/chain-signatures/node/src/mesh/connection.rs
@@ -9,6 +9,7 @@ use crate::protocol::contract::primitives::Participants;
 use crate::protocol::ParticipantInfo;
 use crate::protocol::ProtocolState;
 use crate::web::StateView;
+use mpc_keys::hpke::Ciphered;
 
 // TODO: this is a basic connection pool and does not do most of the work yet. This is
 //       mostly here just to facilitate offline node handling for now.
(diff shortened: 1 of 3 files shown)


ailisp (dismissed):
Looks good to me. How is empty message handled btw? If it's considered an error we may handle it as a proper status checking request

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -186,4 +197,20 @@ impl Pool {
             Err(_) => Err(FetchParticipantError::Timeout),
         }
     }
+
+    async fn test_empty_msg(
```
ChaoticTempest: let's rename this to `send_empty_msg` instead since we should reserve `test_*` functions for testing purposes

ChaoticTempest (dismissed):

ChaoticTempest:
Seems like all the recent PRs are running into this issue:
```
Error: Action #0: CompilationError(PrepareError(Deserialization))
```

I've tried compiling and running tests on a linux machine and they pass so not sure what's going on here. Could be a caching issue

volovyks:
@ailisp can you please clean our cache?

volovyks (dismissed):

ppca:
> Looks good to me. How is empty message handled btw? If it's considered an error we may handle it as a proper status checking request

looking at https://github.com/near/mpc/blob/ecc6fd32ba623025963f34810dca261ec1b5f0d4/chain-signatures/node/src/web/mod.rs#L78, the empty message will just return Ok(()) and not do anything else.

ailisp:
@volovyks done

ChaoticTempest (approved):

PR #902, opened 2024-10-18
fix: prevent messages from taking too long to be sent

ChaoticTempest:
messages will occasionally take over 1s to send, which can stall the node from doing work. This PR makes all the `send_encrypted` be ran concurrently such that the node don't wait until the previous `send_encrypted` gets completed. I've seen the message sending portion go from 1-2s max to 300ms max so this should be a good improvement with no breaking changes

### DIFF
--- a/chain-signatures/node/src/http_client.rs
+++ b/chain-signatures/node/src/http_client.rs
@@ -11,6 +11,8 @@ use std::time::{Duration, Instant};
 use tokio_retry::strategy::{jitter, ExponentialBackoff};
 use tokio_retry::Retry;
 
+use near_account_id::AccountId;
+
 #[derive(Debug, thiserror::Error)]
 pub enum SendError {
     #[error("http request was unsuccessful: {0}")]
@@ -33,7 +35,7 @@ pub enum SendError {
 
 async fn send_encrypted<U: IntoUrl>(
     from: Participant,
-    client: &Client,
+    client: Client,
     url: U,
     message: Vec<Ciphered>,
 ) -> Result<(), SendError> {
@@ -75,13 +77,21 @@ async fn send_encrypted<U: IntoUrl>(
 
 // TODO: add in retry logic either in struct or at call site.
 // TODO: add check for participant list to see if the messages to be sent are still valid.
-#[derive(Default)]
 pub struct MessageQueue {
     deque: VecDeque<(ParticipantInfo, MpcMessage, Instant)>,
     seen_counts: HashSet<String>,
+    account_id: AccountId,
 }
 
 impl MessageQueue {
+    pub fn new(id: &AccountId) -> Self {
+        Self {
+            deque: VecDeque::new(),
+            seen_counts: HashSet::new(),
+            account_id: id.clone(),
+        }
+    }
+
     pub fn len(&self) -> usize {
         self.deque.len()
     }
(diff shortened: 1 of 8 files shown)


chain-signatures/node/src/protocol/mod.rs
```diff
@@ -249,15 +249,17 @@ impl MpcSignProtocol {
...
                 .inc();
+
+            let msg_time = Instant::now();
+            let mut msg_count = 0;
             loop {
                 let msg_result = self.receiver.try_recv();
                 match msg_result {
                     Ok(msg) => {
-                        tracing::debug!("received a new message");
+                        msg_count += 1;
```
ailisp, line 259: Nice, this was too noisy

chain-signatures/node/src/http_client.rs
```diff
@@ -162,54 +166,58 @@ impl MessageQueue {
...
+                }
+                Err(err) => {
+                    tracing::error!(?err, "message queue task failure");
                 }
             }
         }
 
-        if uncompacted > 0 {
+        let elapsed = outer.elapsed();
+        if elapsed > Duration::from_millis(100) && uncompacted > 0 {
```
ailisp: 100ms latency might be too tight. Latency from Hong Kong to LA or Germany to LA is about 150ms. If we can't have all nodes in US, 100ms can't be achieved

ailisp (dismissed):
Nice improvement by concurrently sending messages!

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -266,6 +268,7 @@ impl MpcSignProtocol {
                     }
                 }
             }
+            tracing::debug!("received {msg_count} messages in {:?}", msg_time.elapsed());
```
ppca, line 271: Nice!

chain-signatures/node/src/http_client.rs
ppca: I guess this means if less than 100 ms has passed then sent messages won't be logged? I'd prefer to still see how many messages have been sent if any, even if the time taken is very fast.

ppca (dismissed):
LGTM except the tracing comment.

chain-signatures/node/src/http_client.rs
ChaoticTempest: whoops, that was not supposed to make it to the PR. This was just me debugging locally

chain-signatures/node/src/protocol/mod.rs
volovyks, line 271: nice! this was annoying

chain-signatures/node/src/http_client.rs
```diff
@@ -162,54 +166,57 @@ impl MessageQueue {
...
-                    client,
-                    &info.url,
+                    client.clone(),
+                    info.url.clone(),
                     encrypted_partition,
                     Duration::from_millis(self.message_options.timeout),
-                )
-                .await
-                {
-                    crate::metrics::NUM_SEND_ENCRYPTED_FAILURE
```
volovyks, line 186: Don't we want to keep this metric? MAybe move it to `send_encrypted`.

volovyks (dismissed):
Looks good

chain-signatures/node/src/http_client.rs
ChaoticTempest: it's not used at all right now in our dashboard but I can add it back if there's value in it

volovyks: let's do that, we have only one metric for messages now

PR #904, opened 2024-10-22
fix: contract unit tests

ChaoticTempest:
This should fix all the contract's unit testing. This also commits the `Cargo.lock` file for chain-signatures/contract so hopefully that also fixes integration tests in GH action if stuff is being cached


volovyks (dismissed):

ailisp:
Seems not triggering the contract unit test, see unit/check & test log

ChaoticTempest:
for some reason, we don't have the equivalent commands for running the chain signatures unit tests in CI which is strange. I remember we had one once upon a time but don't know where it went

ppca (approved):

volovyks (approved):

PR #907, opened 2024-10-23
Store Triples using Redis, refactor TripleManager

volovyks:
What was done:
- Removed duplicate storage (DB and memory) approach
- Datastore replaced with Redis
- Refactoring of TripleManager

Other:
- I removed StuckMonitor because it does not fit well into the design with Redis. I assume it was used for testing. I can add a lightweight version in Triple and Presignature managers if requested.


chain-signatures/node/src/protocol/triple.rs
```diff
@@ -96,30 +96,24 @@ pub struct TripleManager {
...
     pub threshold: usize,
     pub epoch: u64,
-    pub triple_storage: LockTripleNodeStorageBox,
     pub my_account_id: AccountId,
 }
 
 impl fmt::Debug for TripleManager {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("TripleManager")
-            .field("triples", &self.triples.keys().collect::<Vec<_>>())
```
volovyks, line 116: Now this can be checked in Redis, with all details

Issue #763, opened 2024-07-30
Alow to request signature for the same payload


volovyks:

volovyks:
@ailisp can you please take a look at this?

volovyks:
Related code snippet: https://github.com/gagdiez/yield-resume/blob/main/contracts/retry-on-error/src/lib.rs

PR #909, opened 2024-10-25
enable batching multiple sign calls in one batch transaction

### LINKED ISSUE #880: Accept multiple signature requests in single transaction
MPC system should accept multiple payloads for signing. Obviously we can accept a longer TX time, and pay more, but it seems like most teams working with Bitcoin UTXOs are scratching their heads trying to come up with solutions for how to spend multiple UTXOs.

Each UTXO to spend requires it's own signature. So with the current configuration of MPC we can really only spend from one UTXO.

Possible solutions outside of the MPC accepting multiple payloads would be:
- chain promise calls from a contract CONS: derived accounts are based on contract, have to make a self call to contract before returning the parallel calls made to MPC back to client
- make multiple calls from client CONS: wallet / user needs to sign multiple TXs

Seeing as how everyone is going to want to do this sooner than later, for Bitcoin UTXOs and other applications, I think it should be a priority.

(requested by @mattlockyer)

### LINKED ISSUE #901: Transaction with multiple actions can create identical requests
### Description

If a transaction is executed with multiple actions to call `sign` with identical payloads, all of these sign calls will pass the check at line:

https://github.com/near/mpc/blob/ecc6fd32ba623025963f34810dca261ec1b5f0d4/chain-signatures/contract/src/lib.rs#L164

Because the contract only adds the request after an async call to `sign_helper`.

If multiple actions are used, then multiple calls to `sign_helper` will be fired, resulting in multiple calls to `env::promise_yield_create` which may go on to create further issues.

See transaction here:
https://testnet.nearblocks.io/txns/FWB6Wq8tYNLFgB3zFnHZ5rmunfEHuVKscowakeenLAMq#execution

ppca:
changes:
1) `mark_request_received(request)` with <request, None> in sign(), so that same payloads in the same batch tx will not be able to go thru. Only one should go thru;
2) use `SignRequestIdentifier {receipt_id, epsilon, payload}` to index sign requests instead of `receipt_id` in backend code. When sending multiple sign() in one batch tx, the receipt_id and entropy of all of them are exactly same. What should differentiate is epsilon (which reflects path + predecessor) or payload;
3) add an integration test_batch_signature with 3 sign calls and verify the signature response in the receipt is correct for each of the payload.


chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -234,7 +234,7 @@ impl PresignatureManager {
             )));
         }
 
-        tracing::info!(id, "starting protocol to generate a new presignature");
+        tracing::debug!(id, "starting protocol to generate a new presignature");
```
volovyks: isn't it important enough for `info`?

chain-signatures/contract/src/lib.rs
```diff
@@ -166,6 +172,7 @@ impl VersionedMpcContract {
                 "sign: predecessor={predecessor}, payload={payload:?}, path={path:?}, key_version={key_version}",
             );
             env::log_str(&serde_json::to_string(&near_sdk::env::random_seed_array()).unwrap());
+            self.mark_request_received(&request);
```
volovyks, line 175: I assume a duplicate request with the same payload, epsilon, and crypto hash will not pass the check on line 170 and throw an error on 184. Let's have a separate test for that (batch transaction with 2+ requests with the same payload and path). And it should not break the counter.

BTW, the error should not be "PayloadCollision" anymore. Should be SignRequestCollision.

chain-signatures/node/src/protocol/message.rs
```diff
@@ -103,7 +104,7 @@ pub struct MpcMessageQueue {
     resharing_bins: HashMap<u64, VecDeque<ResharingMessage>>,
     triple_bins: HashMap<u64, HashMap<TripleId, VecDeque<TripleMessage>>>,
     presignature_bins: HashMap<u64, HashMap<PresignatureId, VecDeque<PresignatureMessage>>>,
-    signature_bins: HashMap<u64, HashMap<CryptoHash, VecDeque<SignatureMessage>>>,
+    signature_bins: HashMap<u64, HashMap<SignRequestIdentifier, VecDeque<SignatureMessage>>>,
```
volovyks, line 106: Since we are preparing the integration with Ethereum, we should not propagate the ReceiptId type to the MPC layer, since it's NEAR specific. [u8; 32] is an option, since it's a type wrapper around it anyway.

volovyks (commented):
Looks good overall

This PR can close https://github.com/near/mpc/issues/763 too. If so, let's add a separate test.

chain-signatures/node/src/protocol/message.rs
ppca: Do we want 
```

pub struct SignRequestIdentifier {
    pub receipt_id: [u8; 32],
    pub epsilon: Vec<u8>,
    pub payload: Vec<u8>,
}
```
or
```

pub struct SignRequestIdentifier {
    pub receipt_id: Vec<u8>,
    pub epsilon: Vec<u8>,
    pub payload: Vec<u8>,
}
```
?

chain-signatures/contract/src/lib.rs
ppca: Done. Added test_batch_duplicate_signature where 2 identical requests are sent. The test checks that tx fails with RequestCollision.

chain-signatures/node/src/protocol/message.rs
volovyks: I think fixed length is better.

chain-signatures/contract/src/lib.rs
ChaoticTempest: hmm, I'm against changing the name, because it's going to be a breaking change now if somebody was directly matching against the name of the error

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -165,7 +149,7 @@ pub struct SignatureGenerator {
     pub presignature_id: PresignatureId,
     pub request: ContractSignRequest,
     pub epsilon: Scalar,
-    pub receipt_id: CryptoHash,
+    pub receipt_id: [u8; 32],
```
ChaoticTempest: why change these to non-`CryptoHash` type?

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -227,18 +211,35 @@ pub struct GenerationRequest {
...
+
+impl SignRequestIdentifier {
+    pub fn new(receipt_id: [u8; 32], epsilon: Scalar, payload: Scalar) -> Self {
+        Self {
+            receipt_id,
+            epsilon: borsh::to_vec(&SerializableScalar { scalar: epsilon }).unwrap(),
+            payload: borsh::to_vec(&SerializableScalar { scalar: payload }).unwrap(),
+        }
+    }
+}
```
ChaoticTempest: why are we making epsilon and the payload vectors? Is it for hashing? If so, just implement hashing for `SerializableScalar`

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -356,13 +357,13 @@ impl SignatureManager {
     #[allow(clippy::result_large_err)]
     fn retry_failed_generation(
         &mut self,
-        receipt_id: ReceiptId,
+        sign_request_identifier: SignRequestIdentifier,
```
ChaoticTempest, line 360: this is quite long now as a variable name. Let's just simplify all of them to just `request_id: SignRequestIdentifier`

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: ok, so talked with @volovyks about this one and it seems like we're moving to a more generic naming for these because other chains don't have this. We can change it to some other type later

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -227,18 +211,35 @@ pub struct GenerationRequest {
...
     pub epsilon: Scalar,
-    pub receipt_id: CryptoHash,
+    pub receipt_id: [u8; 32],
     pub entropy: [u8; 32],
     pub sign_request_timestamp: Instant,
 }
 
+#[derive(Debug, Clone, Eq, Hash, PartialEq)]
+pub struct SignRequestIdentifier {
+    pub receipt_id: [u8; 32],
```
volovyks: receipt is still in the name here

chain-signatures/node/src/protocol/signature.rs
volovyks: Have you tried using the newly introduced `SignRequestIdentifier` everywhere?

chain-signatures/node/src/protocol/signature.rs
ppca: I'll do all the rafactoring in an upcoming PR. Want to prioritize making batch multiple sign possible on dev first.

chain-signatures/node/src/protocol/signature.rs
ppca: This will come in a refactoring PR

ppca:
Talked with @volovyks this morning, since many people have been waiting for batch tx with multiple signs to work on dev, I want to prioritize merging this fix, then in the next PR refactor the code. @ChaoticTempest

chain-signatures/node/src/protocol/signature.rs
ppca: removed receipt_id field from everywhere but indexer

volovyks (approved):
Let's do follow up PR later and fix the issue now.

PR #910, opened 2024-10-26
fix: Smart contract panicked: Callback computation 0 was not successful

### LINKED ISSUE #870: Smart contract panicked: Callback computation 0 was not successful
https://nearblocks.io/txns/GJViX5rAWhWLCZ2fn29EWMWaCuBTgbLzjXhxn6pNUDHH#

ppca:
problem was prepaid gas for clear_state_on_finish was too low. 
I doubled prepaid gas for both clear_state_on_finish and return_signature_on_finish

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -41,10 +41,10 @@ const GAS_FOR_SIGN_CALL: Gas = Gas::from_tgas(50);
 const DATA_ID_REGISTER: u64 = 0;
 
 // Prepaid gas for a `clear_state_on_finish` call
-const CLEAR_STATE_ON_FINISH_CALL_GAS: Gas = Gas::from_tgas(10);
+const CLEAR_STATE_ON_FINISH_CALL_GAS: Gas = Gas::from_tgas(20);
 
 // Prepaid gas for a `return_signature_on_finish` call
-const RETURN_SIGNATURE_ON_FINISH_CALL_GAS: Gas = Gas::from_tgas(5);
+const RETURN_SIGNATURE_ON_FINISH_CALL_GAS: Gas = Gas::from_tgas(10);
 
 // Prepaid gas for a `update_config` call
 const UPDATE_CONFIG_GAS: Gas = Gas::from_tgas(5);


volovyks:
Gas consumption must be identical on all calls. If it's enough for one it should be enough for any call.
We are hitting an edge case where we consume more gas than expected, like when people submit multiple `sign()` call actions in one transaction.

ppca:
> Gas consumption must be identical on all calls. If it's enough for one it should be enough for any call. We are hitting an edge case where we consume more gas than expected, like when people submit multiple `sign()` call actions in one transaction.

More explanation in this issue: https://github.com/near/mpc/issues/870
Basically not all calls will consume the same gas. For example, sometimes clear_state_on_finish need to wait many blocks for the return value, so it can cost more gas when the waiting is longer. 
And when I look into this issue, indeed it fails with `callback computation 0 was not successful` when `clear_state_on_finish` has consumed more gas than normally. 
So what it means is, clear_state_on_finish needs more gas when waiting is longer. 

The change I made is giving the `clear_state_on_finish` call more prepaid gas. This will not change the total gas we require for the sign call, it is adjusting how much goes to clear_state_on_finish. I also increased the gas prepaid for `return_signature_on_finish`, in order to reserve some more buffer there in case that call also consumes more gas when waiting is longer.

volovyks (approved):

ChaoticTempest:
@ppca we should probably base it on the exact number then instead of purely doubling it everytime we hit the limit. 200 blocks is the limit it should be waiting for anyways, so we can have a definite constant here.

We should also do a check for `prepaid_gas < GAS_FOR_SIGN_CALL + CLEAR_STATE_ON_FINISH_CALL_GAS + RETURN_SIGNATURE_ON_FINISH_CALL_GAS` because now `GAS_FOR_SIGN_CALL` amount might not be enough or adjust the constant to make it enough

ppca:
> @ppca we should probably base it on the exact number then instead of purely doubling it everytime we hit the limit. 200 blocks is the limit it should be waiting for anyways, so we can have a definite constant here.
> 
> We should also do a check for `prepaid_gas < GAS_FOR_SIGN_CALL + CLEAR_STATE_ON_FINISH_CALL_GAS + RETURN_SIGNATURE_ON_FINISH_CALL_GAS` because now `GAS_FOR_SIGN_CALL` amount might not be enough or adjust the constant to make it enough

Yeah good idea to base it on exact number. Let's ask Saketh and we'll update the number.

PR #914, opened 2024-10-30
fix: presignature storage locks

ChaoticTempest:
The mutability and lock acquisition was being done incorrectly, where multiple write locks were trying to be acquired


volovyks (approved):
Thank you!

PR #927, opened 2024-11-06
fix: include complete timeout in signature generator failure

ppca:


volovyks (approved):

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -533,6 +533,9 @@ impl SignatureManager {
                                 ));
                             } else {
                                 self.completed.insert(sign_request_identifier.clone(), Instant::now());
+                                crate::metrics::SIGNATURE_GENERATOR_FAILURES
```
volovyks, line 536: @ppca on a second thought, I'm not sure if this is what we need. Both SIGNATURE_GENERATOR_FAILURES and SIGNATURE_FAILURES are barely matching their naming.

